PR #25, opened 2022-12-08
deduplicate flakes

Mic92:
without this users end up with multiple copies of nixpkgs, which cannot be overriden from the outside (follows only works on 1 level).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -105,7 +105,7 @@
 
       nixosModules.lanzaboote = { pkgs, lib, ... }: {
         imports = [ ./nix/lanzaboote.nix ];
-        boot.lanzaboote.package = lib.mkDefault self.packages.${pkgs.system}.lanzaboote;
+        boot.lanzaboote.package = lib.mkDefault self.packages.${pkgs.system}.lanzatool;
       };
 
       packages.x86_64-linux = {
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -95,7 +103,10 @@
         inherit lanzatool;
       };
 
-      nixosModules.lanzaboote = import ./nix/lanzaboote.nix;
+      nixosModules.lanzaboote = { pkgs, lib, ... }: {
+        imports = [ ./nix/lanzaboote.nix ];
+        boot.lanzaboote.package = lib.mkDefault self.packages.${pkgs.system}.lanzaboote;
```
blitz, line 108: The naming is a bit awkward, but this needs to be `lanzatool` to avoid breaking the tests:

```suggestion
        boot.lanzaboote.package = lib.mkDefault self.packages.${pkgs.system}.lanzatool;
```

Are you overriding this here so people using the module don't have to use the overlay?

blitz (changes requested):

flake.nix
Mic92: Yes. Overlays are pretty bad from a performance perspective (at least 500MB RAM for evaluation). Also they are harder to understand and not needed here. In most cases they actually shouldn't be used (i.e. when adding new attributes to nixpkgs)

blitz (approved):

blitz:
@Mic92 Thank you!

PR #40, opened 2022-12-26
Move lanzatool to inputsFrom in devShell

nikstur:
When lanzatool is in the packages attr of the devShell, any compliation error in lanzatool means direnv cannot load the environment anymore. Then LSP support in your editor and even cargo in your shell is missing.


RaitoBezarius (approved):
LGTM

PR #41, opened 2022-12-28
Avoid Path -> String Conversions

blitz:
Lanzatool had a couple of avoidable `Path` to `String` conversions. These are problematic, because they can easily fail if a user has a messed up way of calling files. :)

All of them happened in order to pass paths as command arguments. In this case, these conversions are avoidable, because `Command` takes `OsString` as well. And we can construct `OsString` from `Path` without the possibility of failing.


nikstur (approved):
This lets us avoid some ugly hacks, but forces us to do some other ugly hacks. But since it overall decreases lines of code. I'm happy to merge it.

PR #54, opened 2023-01-07
Enforce Nix Style

blitz:
Building on top of #52, this PR enforces nixpkgs-fmt style on our Nix files via pre-commit. This is also checked via CI.

Please review commit-by-commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 result*
 /.direnv
 target
+/.pre-commit-config.yaml
--- a/flake.lock
+++ b/flake.lock
@@ -59,6 +59,22 @@
         "type": "github"
       }
     },
+    "flake-compat_3": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1668681692,
+        "narHash": "sha256-Ht91NGdewz8IQLtWZ9LCeNXMSXHUss+9COoqu6JLmXU=",
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "rev": "009399224d5e398d03b22badca40a37ac85412a1",
+        "type": "github"
+      },
+      "original": {
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "type": "github"
+      }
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
 result*
 /.direnv
 target
+/.pre-commit-config.yaml
```
nikstur, line 4:
```suggestion
.pre-commit-config.yaml
```
I would change it, so that even when you work in a subdirectory (e.g. rust/lanzatool), the .pre-commit file is ignored. I've had this issue before.

nikstur (changes requested):
I think your commits are a little mangled: 4f14ca01973df7d70fd3fc96b32a305873fbf869 and 74182c199a6f95d706ba8cd9d328b41c1b4419a6 do not look right.

.gitignore
blitz: Do you have a specific scenario? Does pre-commit place configs in other directories?

blitz:
> I think your commits are a little mangled: [4f14ca0](https://github.com/nix-community/lanzaboote/commit/4f14ca01973df7d70fd3fc96b32a305873fbf869) and [74182c1](https://github.com/nix-community/lanzaboote/commit/74182c199a6f95d706ba8cd9d328b41c1b4419a6) do not look right.

They are like this on purpose. The first patch contains the semantic change of the pre-commit config, the second is a pure reformat. This makes it easier to handle conflicts and also see what has changed in the history. I wouldn't want to lump together the changed config with the reformat of the whole file.

What would have helped you to understand my intention better?

.gitignore
nikstur: It has happened in the past, but I can't reproduce it right now. So it should be fine.

nikstur:
Sorry, that was just brain lag, I read the commits in the order `git log` would put them and then of course they don't make sense.

nikstur (approved):

PR #57, opened 2023-01-09
Add sbctl to path when lanzaboote is enabled

blitz:
This is useful when debugging or troubleshooting issues. Just as `bootctl` is available, we should make `sbctl` available to give the user the ability to debug things.

### DIFF
--- a/nix/modules/lanzaboote.nix
+++ b/nix/modules/lanzaboote.nix
@@ -45,9 +45,13 @@ in
   };
 
   config = mkIf cfg.enable {
-    boot.bootspec = {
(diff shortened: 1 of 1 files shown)


nikstur:
Is this really something the module should do/decide? Shouldn't this be user configuration?

blitz:
> Is this really something the module should do/decide? Shouldn't this be user configuration?

I'm not sure. Maybe. At the moment, it's useful, because you need the tool anyhow to debug things. But then you already need the tool _before_ you can enable the module, so it's not that useful to include it I guess.

Issue #60, opened 2023-01-12
Improve subproject/crate naming


nikstur:
Lanzaboote is the name of the entire project, but also of the boot stub. I think we should find a better distinction. 

I propose to rename the stub:

Rename the source code directory:  `rust/lanzaboote` -> `rust/stub`
Rename crate in `Cargo.toml`: `lanzaboote` -> `lanzaboote_stub`

To make it even more streamlined we could also rename `lanzatool` 

Rename the source code directory:  `rust/lanzatool` -> `rust/tool`
Rename crate in `Cargo.toml`: `lanzatool` -> `lanzaboote_tool`
Use `lzbt` as the command name.

PR #62, opened 2023-01-13
Rename subprojects

nikstur:
This PR implements the ideas from #60.

It is split into two commits. The first only renames the directories, while the second changes names inside the various files.


RaitoBezarius (approved):

blitz (changes requested):
Nice! Can you make a pass through the README and the quick start docs and update the names there as well?

nikstur:
I think I got everything.

blitz (approved):

PR #63, opened 2023-01-14
blake3 -> sha256

nikstur:
Using the sha2 crate instead of blake3 decreases the binary size of the stub by around 50%. 
From ~230K to ~90K

### DIFF
--- a/rust/lanzaboote/Cargo.lock
+++ b/rust/lanzaboote/Cargo.lock
@@ -3,18 +3,6 @@
 version = 3
 
 [[package]]
-name = "arrayref"
-version = "0.3.6"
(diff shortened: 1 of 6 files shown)


RaitoBezarius (approved):
No strong opinion on this PR.

nikstur:
@blitz any opinion on this? I'm not convinced myself that this implementation is better. Although I like that the stub size decreases considerably.

blitz:
Didn't have time to look into the diff but if the stub size shrinks I'm all for it. 👍

nikstur:
This is easy enough to revert, so I'll merge it.

PR #69, opened 2023-01-17
Refactor tempdir: Add SecureTempDirExt

nikstur:
Depends on #62 

Add an extension to TempDir that allows to create secure tempfiles. This
way, everything related to creating secure tempfiles is bundled in a
single place and can easily be reused.

### DIFF
--- a/rust/tool/src/install.rs
+++ b/rust/tool/src/install.rs
@@ -5,7 +5,6 @@ use std::process::Command;
 
 use anyhow::{Context, Result};
 use nix::unistd::sync;
-use tempfile::tempdir;
 
 use crate::esp::EspPaths;
 use crate::gc::Roots;
@@ -13,6 +12,7 @@ use crate::generation::{Generation, GenerationLink};
 use crate::os_release::OsRelease;
 use crate::pe;
 use crate::signature::KeyPair;
+use crate::utils::SecureTempDirExt;
 
 pub struct Installer {
     gc_roots: Roots,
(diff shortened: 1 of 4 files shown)


rust/tool/src/install.rs
```diff
@@ -121,23 +121,18 @@ impl Installer {
...
             assemble_kernel_cmdline(&bootspec.init, bootspec.kernel_params.clone());
 
-        // prepare a secure temporary directory
-        // permission bits are not set, because when files below
-        // are opened, they are opened with 600 mode bits.
-        // hence, they cannot be read except by the current user
-        // which is assumed to be root in most cases.
-        // TODO(Raito): prove to niksnur this is actually acceptable.
-        let secure_temp_dir = tempdir()?;
+        // This tempdir must live for the entire liftime of the current function.
```
blitz, line 124:
```suggestion
        // This tempdir must live for the entire lifetime of the current function.
```

rust/tool/src/utils.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use anyhow::{Context, Result};
+use tempfile::TempDir;
+
+/// A temporary directory that enables creating secure temporary files in it.
+///
+/// The temporary directory has three useful properties:
+///
+/// - Files are created with mode 0o600, so that they are only accessible by the current user.
+/// - Files are named and not entirely ephemeral (unlike a real temporary file).
```
blitz, line 14:
```suggestion
/// - Files are named and not ephemeral (unlike a real temporary file).
```

rust/tool/src/utils.rs
```diff
@@ -0,0 +1,55 @@
...
+/// - The directory and its children are cleaned up (i.e. deleted) when the variable that holds the
+///   directory goes out of scope.
+///
+/// This lightly protects against an attacker without root access from modifying files undetected.
```
blitz, line 18: So does it protect or not? If not, what does it not protect against?

rust/tool/src/utils.rs
```diff
@@ -0,0 +1,55 @@
...
+/// Additionally, because the files have named paths, they can be passed to external programs while
+/// still being securely deleted after they are not needed anymore.
+pub struct SecureTempDir(TempDir);
```
blitz, line 21: nit: You are extending a type with some methods. The typical way in Rust to achieve this is not to wrap the type in a struct, because this will lead to "pass-through" implementations of some functions (your `path` is a case in point). Instead you define a trait with your new functionality and implement it for `TempDir`: https://cmcenroe.me/2016/08/22/rust-extending-type.html

blitz (approved):
I've left some nits. Feel free to ignore.

For the future, it would be nice to have at least a one-liner what your refactoring tries to achieve in either the commit message or the PR description. We should get a good example for any future contributors. :)

rust/tool/src/utils.rs
nikstur: Thanks! I turned it into an extension trait.

rust/tool/src/utils.rs
nikstur: I clarified this a tiny bit. For more detail, I'd have to do some more extensive threat modeling.

blitz (approved):

PR #71, opened 2023-01-21
Enable Typo Checks

blitz:
None of us are native English speakers. This PR enables automatic typo checks in our tree. I've also fixed all typos that it found.


nikstur (approved):
This is a good idea, regardless of how good our english is ;)

PR #74, opened 2023-01-22
tool: split esp paths

nikstur:
To access paths on the ESP before or after installing generations, split EspPaths into general EspPaths that only depend on the path to the ESP and EspGenerationPaths which additionally depend on generation specific information (e.g. version number and initrd filename).

This refactoring is in preparation to a larger change to the way systemd-boot is installed.

### DIFF
--- a/rust/tool/src/esp.rs
+++ b/rust/tool/src/esp.rs
@@ -5,14 +5,12 @@ use anyhow::{Context, Result};
 
 use crate::generation::Generation;
 
+/// Paths to the boot files that are not specific to a generation.
 pub struct EspPaths {
     pub esp: PathBuf,
     pub efi: PathBuf,
     pub nixos: PathBuf,
-    pub kernel: PathBuf,
-    pub initrd: PathBuf,
     pub linux: PathBuf,
-    pub lanzaboote_image: PathBuf,
     pub efi_fallback_dir: PathBuf,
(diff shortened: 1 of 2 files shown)


rust/tool/src/esp.rs
```diff
@@ -70,6 +56,38 @@ impl EspPaths {
     }
 }
 
+/// Contains the generation-specific paths to the ESP.
```
blitz, line 59: nit: I honestly had a bit of trouble making sense of this. Maybe:

```
/// Paths to the boot files of a specific generation.
///
/// All of these paths point into the ESP.
```

rust/tool/src/esp.rs
```diff
@@ -5,62 +5,48 @@ use anyhow::{Context, Result};
 
 use crate::generation::Generation;
 
+/// Contains generation-independent paths on the ESP.
```
blitz, line 8: nit: Maybe:

```
/// Path to boot files that are not specific to a generation.
```

blitz (approved):
Looks pretty straightforward. Feel free to ignore the nits.

rust/tool/src/esp.rs
nikstur: I still have a hard time finding a good comment/explanation. Your suggestions are definetely an improvement though.

PR #76, opened 2023-01-23
tool: smarter systemd-boot install 

nikstur:
The process of installing systemd-boot is "smarter" because it now
considers a a few conditions instead of doing nothing if there is a file
at the deistination path. systemd-boot is now forcibly installed (i.e.
overwriting any file at the destination) if (1) there is no file at the
destination, OR (2) a newer version of systemd-boot is available, OR (3)
the signature of the file at the destination could not be verified.

Fixes part of #39 

Will produce a small and easy to fix merge conflict with #75

### DIFF
--- a/rust/tool/src/install.rs
+++ b/rust/tool/src/install.rs
@@ -1,7 +1,9 @@
+use std::ffi::CStr;
 use std::fs;
 use std::os::unix::prelude::PermissionsExt;
 use std::path::{Path, PathBuf};
 use std::process::Command;
+use std::str::FromStr;
 
 use anyhow::{Context, Result};
 use nix::unistd::sync;
@@ -17,6 +19,7 @@ use crate::utils::SecureTempDirExt;
 pub struct Installer {
     gc_roots: Roots,
     lanzaboote_stub: PathBuf,
+    systemd: PathBuf,
     key_pair: KeyPair,
     configuration_limit: usize,
     esp_paths: EspPaths,
@@ -26,6 +29,7 @@ pub struct Installer {
 impl Installer {
     pub fn new(
         lanzaboote_stub: PathBuf,
+        systemd: PathBuf,
         key_pair: KeyPair,
         configuration_limit: usize,
         esp: PathBuf,
@@ -38,6 +42,7 @@ impl Installer {
         Self {
             gc_roots,
             lanzaboote_stub,
+            systemd,
             key_pair,
             configuration_limit,
             esp_paths,
@@ -65,6 +70,8 @@ impl Installer {
                 .collect()
         };
         self.install_links(links)?;
+
+        self.install_systemd_boot()?;
 
         // Only collect garbage in these two directories. This way, no files that do not belong to
         // the NixOS installation are deleted. Lanzatool takes full control over the esp/EFI/nixos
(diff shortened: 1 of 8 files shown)


rust/tool/src/install.rs
```diff
@@ -281,3 +316,56 @@ fn ensure_parent_dir(path: &Path) {
...
+        _ => return Ok(true),
+    };
+
+    Ok(from_version > to_version)
+}
+
+/// Read the version of a systemd-boot binary from its `.osrel` section.
+///
+/// The version is parsed into a f32 because systemd does not follow strict semver conventions. A
+/// f32, however, should parse all systemd versions and enables usefully comparing them.
```
blitz, line 349: nit: It is a bit of a hack though. ;)

rust/tool/src/os_release.rs
```diff
@@ -25,17 +25,36 @@ impl OsRelease {
...
             generation
                 .describe()
                 .context("Failed to describe generation.")?,
         );
 
         Ok(Self(map))
     }
+
+    /// Parses the string representation of a os-release file.
+    pub fn from_str(value: &str) -> Result<Self> {
```
blitz, line 41: This doesn't handle entries with quotes correctly. (See also the [spec](https://www.freedesktop.org/software/systemd/man/os-release.html).) They sort of work by accident, because you consider the quotes part of the value and then write them back out.

Maybe something like this would help: https://crates.io/crates/os-release

rust/tool/src/install.rs
nikstur: Oh yes big time. Do you have a better solution? Or should I just make this clear in the comment as well?

rust/tool/src/pe.rs
```diff
@@ -180,6 +180,22 @@ fn file_size(path: impl AsRef<Path>) -> Result<u64> {
         .size())
 }
 
+/// Read the data within a section of a PE binary at the provided path.
```
blitz, line 183: But you don't provide a path as parameter. The comment seems stale.

blitz (changes requested):
This code was really easy to follow thanks to the useful comments. I love it.

My only concern here is the yolo parsing/printing of the os-release file. If we can use one of the existing parsing libraries that also handles the quoting correctly, I would sleep more calmly. :)

rust/tool/src/os_release.rs
nikstur: I looked at the os-release crate because I had the same instinct of avoiding yolo parsing, but sadly it's (1) unmaintained since 2018 and (2) actually performs the same hack (just slightly less elegantly): Compare [mine](https://github.com/nix-community/lanzaboote/blob/smart-systemd-boot-install/rust/tool/src/os_release.rs#L53) with [theirs](https://github.com/pop-os/os-release/blob/master/src/lib.rs#L37)

We actually never write out a file we have previously read. So I'm not too worried about yolo parsing from that perspective. 
I should add a comment, however, that writing out a previously read file might not produce the same representation.

blitz: Fair point. I've opened #77 to fix this later.

rust/tool/src/install.rs
blitz: If you make it clear in the comments this is fine for me. It does the job.

If someone wants to they can just add a small parser from `String` to `(u32, u32)`.

blitz (approved):

rust/tool/src/install.rs
nikstur: I'll do this in a separate PR.

PR #78, opened 2023-01-25
tool: make systemd version parsing robust

nikstur:
To make handling systemd versions more robust, they are parsed into a u32 tuple instead of an f32. Additionally, some unit tests for correct parsing and comparing of versions are added.

### DIFF
--- a/rust/tool/src/systemd.rs
+++ b/rust/tool/src/systemd.rs
@@ -1,0 +1,98 @@
+use std::ffi::CStr;
+use std::fs;
+use std::path::Path;
+use std::str::FromStr;
+
+use anyhow::{Context, Result};
+
+use crate::os_release::OsRelease;
+use crate::pe;
+
+/// A systemd version.
+///
+/// The version is parsed into a u32 tuple because systemd does not follow strict semver
+/// conventions. A major version without a minor version, e.g. "252" is represented as `(252, 0)`.
+#[derive(PartialEq, PartialOrd, Eq, Debug)]
+pub struct SystemdVersion(u32, u32);
+
+impl SystemdVersion {
+    /// Read the systemd version from the `.osrel` section of a systemd-boot binary.
+    pub fn from_systemd_boot_binary(path: &Path) -> Result<Self> {
+        let file_data = fs::read(path).with_context(|| format!("Failed to read file {path:?}"))?;
+        let section_data = pe::read_section_data(&file_data, ".osrel")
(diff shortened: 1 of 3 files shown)


rust/tool/src/systemd.rs
```diff
@@ -0,0 +1,84 @@
...
+use anyhow::{Context, Result};
+
+use crate::os_release::OsRelease;
+use crate::pe;
+
+/// A systemd version.
+///
+/// The version is parsed into a u32 tuple because systemd does not follow strict semver
+/// conventions. A major version without a minor version, e.g. "252" is represented as `(252, 0)`.
+#[derive(PartialEq, PartialOrd, Debug)]
```
blitz, line 15: nit: This can also derive `Eq`.

rust/tool/src/systemd.rs
```diff
@@ -0,0 +1,84 @@
...
+            .0
+            .get("VERSION")
+            .context("Failed to extract VERSION key from: {os_release:#?}")?;
+
+        Self::from_str(version_str)
+    }
+
+    /// Parse a systemd version from a string.
+    pub fn from_str(s: &str) -> Result<Self> {
+        let mut split_version_string = s.split('.');
```
blitz, line 46: Couldn't this by `s.split(...)..map(... u32::from_str ...).collect::<Result<Vector<u32>>()?;` 

Then you can also check whether you actually get 2 elements and fix as required. If the version number tends to have crap at the end you can also add a `.take(2)` above.

rust/tool/src/systemd.rs
```diff
@@ -0,0 +1,84 @@
...
+        assert_eq!(parse_version("253"), SystemdVersion(253, 0));
+        assert_eq!(parse_version("252.4"), SystemdVersion(252, 4));
+        assert_eq!(parse_version("251.11"), SystemdVersion(251, 11));
+    }
+
+    #[test]
+    fn compare_version_correctly() {
+        assert!(parse_version("253") > parse_version("252"));
+        assert!(parse_version("253") > parse_version("252.4"));
+        assert!(parse_version("251.8") == parse_version("251.8"));
```
blitz, line 78: It would be nice to add a negative test here as well with some "213;k;13" or "-1.3.123" or "".

blitz: In `parse_version_correctly` not in the compare test obviously. :)

blitz (approved):
Looks good to me. I've left some ideas for improvement, but since it is already an improvement over the status quo, I'm also fine with merging as is.

PR #82, opened 2023-01-28
nix.tests: clean up

nikstur:
The test attributes and names are simplified and standardized. They now roughly follow the same structure as the systemd-boot test in Nixpkgs. Some comments are added and variable names changed to make it more clear what they actually do.


RaitoBezarius (approved):
Thank you for taking care of this!

PR #90, opened 2023-02-02
Fix CI Regression after Linux Loader Fix

blitz:
Now that we don't sign the kernel anymore, we need to manually invalidate its checksum.

This is a regression from #75. See https://github.com/nix-community/lanzaboote/pull/75#issuecomment-1413953437.


RaitoBezarius (approved):
LGTM.

PR #101, opened 2023-02-08
Improve generation name

### LINKED ISSUE #99: Generation name in boot screen contains twice NixOS and kernel version
An systemd boot entry looks like: ``NixOS (Generation 85 NixOS 23.05, Linux Kernel 6.1.9, Built on 2023-02-08)``
a lanzaboote one is a bit longer and contains redundant information: ``NixOS Stoat 23.05 (Linux 6.1.9) (Generation 85 NixOS 23.05), Linux Kernel 6.1.9, Built on 1970-01-01)``

I don't mind the release name (Stoat), that is a nice feature we should maybe add to systemd-boot in nixos, too but the first Linux version can go. Maybe the first release number but not sure on that one.

nikstur:
Leverage the bootspec `label` field in its intended way. The VERSION_ID
of the os-release in the stub now only contains the generation number
and the build time. This makes a correct PRETTY_NAME entirely dependent
on correct information in the bootspec `label` field.

This should yield a display name similar to this: `NixOS Stoat 23.05 (Linux 6.1.9) (Generation 85, Built on 2023-02-01)`

Depends on #100

Closes #99


blitz:
@nikstur  Looks good. Can you rebase?

PR #102, opened 2023-02-09
nixos/lanzaboote: use `config.systemd.package`

ercao:
Signed-off-by: ercao <vip@ercao.cn>


RaitoBezarius (approved):

blitz:
I ran the CI checks manually. Thanks for the fix!

PR #107, opened 2023-02-20
tool: use random names for secure tempfiles

nikstur:

#108 -> (#107) -> #110 -> #109

Using random names for tempfiles makes handling them easier. It reduces the amount of noise in the code because no custom name needs to be provided for each tempfile. The names were not really useful in any case.

It also does not burden the developer with ensuring uniqueness of names. This is relevant when files for multiple generations need to be stored in the same directory (e.g. because they need to be accessed after handling one generation).

Out of an abundance of caution, 32 random alphanumeric characters are chosen for each filename. The tempfile crate, in comparison, only chooses 8. 32 characters should be enough to avoid collisions, even if the PRNG is not of cryptographic quality.

Depens on #108

### DIFF
--- a/rust/tool/src/utils.rs
+++ b/rust/tool/src/utils.rs
@@ -1,15 +1,20 @@
+use std::ffi::OsString;
 use std::fs;
 use std::io::Write;
+use std::iter::repeat_with;
 use std::os::unix::fs::OpenOptionsExt;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use anyhow::{Context, Result};
 use tempfile::TempDir;
 
+/// The number of random alphanumeric characters in the tempfiles.
+const TEMPFILE_RANDOM_LENGTH: usize = 32;
+
 /// Extension for a temporary directory that enables creating secure temporary files in it.
 pub trait SecureTempDirExt {
-    fn create_secure_file(&self, file_name: &str) -> Result<fs::File>;
-    fn write_secure_file(&self, file_name: &str, contents: impl AsRef<[u8]>) -> Result<PathBuf>;
+    fn create_secure_file(&self, path: &Path) -> Result<fs::File>;
+    fn write_secure_file(&self, contents: impl AsRef<[u8]>) -> Result<PathBuf>;
 }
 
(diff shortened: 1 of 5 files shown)


rust/tool/src/utils.rs
```diff
@@ -47,3 +51,16 @@ impl SecureTempDirExt for TempDir {
...
 }
+
+/// Generate a random name for a temporary file.
+///
+/// This is heavily inspired by the way temporary names are generated in the `tempfile` crate.
+/// Since the `tempfile` crate does not expose this functionality, we have to recreate it here.
+pub fn tmpname() -> OsString {
+    let mut buf = OsString::with_capacity(TEMPFILE_RANDOM_LENGTH);
+    let mut char_buf = [0u8; 4];
+    for c in repeat_with(fastrand::alphanumeric).take(TEMPFILE_RANDOM_LENGTH) {
```
blitz, line 62: This is used in `write_secure_file` and the second line in the `fastrand` [docs](https://docs.rs/fastrand/1.9.0/fastrand/) says: "The implementation uses [Wyrand](https://github.com/wangyi-fudan/wyhash), a simple and fast generator but not cryptographically secure."

In our usecase this is not important, but I would at least mention this in a comment or the `tmpfile` docstring.

**That being said**, aren't you reimplementing [`NamedTempFile::new_in`](https://docs.rs/tempfile/3.3.0/tempfile/struct.NamedTempFile.html#method.new_in) and then calling `keep()` on it? This would avoid adding yet another dependency.

blitz (changes requested):

rust/tool/src/utils.rs
nikstur: I wish it was that easy, but `NamedTempFile` already creates a file. So to implement our "secure" tempfiles, we would need to chmod this tempfile after having created it. Additionally, sometimes we actually only need a path without creating the file ourselves (e.g. when calling objcopy). That's why `tmpname()` is `pub`

blitz: Mmmh. Too bad.

blitz (approved):

PR #116, opened 2023-02-25
Make File Installation Order Deterministic

blitz:
Due to the use of hash maps, the order of file installation was not deterministic. I've changed the code the use BTreeMaps instead, which makes this deterministic. While I was here, I tried to simplify the code a bit.

Reported-by: @RaitoBezarius

### DIFF
--- a/rust/tool/src/install.rs
+++ b/rust/tool/src/install.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 use std::fs;
 use std::os::unix::prelude::PermissionsExt;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


rust/tool/src/install.rs
```diff
@@ -342,46 +360,52 @@ struct GenerationArtifacts {
...
-            install_signed(key_pair, from, to)
-                .with_context(|| format!("Failed to sign and install from {from:?} to {to:?}"))?;
-        }
-
-        for (to, from) in &self.unsigned_files {
-            install(from, to)
-                .with_context(|| format!("Failed to install from {from:?} to {to:?}"))?;
-        }
-        Ok(())
+        self.files.iter().try_for_each(|(to, from)| -> Result<()> {
```
nikstur, line 401: nit: I think, it is generally more idiomatic to use for loops when you only rely on side effects in the iteration. See the comment in https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each

nikstur (approved):
Thanks for the fix!

rust/tool/src/install.rs
blitz: Fixed.

PR #121, opened 2023-02-26
tool: init configurable logging

### LINKED ISSUE #55: Ugly backtraces for non-bootspec generations
Lanzatool outputs some intimidating backtraces when it encounters non-bootspec generations. This should be handled more gracefully by having an actionable log messages. Something like "Failed to sign generation X, because it is not bootspec compliant."

```
Installing generation 351
Appending secrets to initrd...
/boot/EFI/BOOT/BOOTX64.EFI already exists, skipping...
/boot/EFI/systemd/systemd-bootx64.efi already exists, skipping...
/boot/EFI/nixos/0n01vj3mq06pc31i2yhxndvhv4kwl2vp-linux-6.1.3-bzImage.efi already exists, skipping...
/boot/EFI/nixos/lzxjx9ykfzhv5c2r4v36v2w7b69yrpc1-initrd-linux-6.1.3-initrd.efi already exists, skipping...
Signing and installing /boot/EFI/Linux/nixos-generation-351.efi...
Successfully installed lanzaboote to '/boot'
Malformed generation: Failed to build generation from link: GenerationLink { version: 350, path: "/nix/var/nix/profiles/system-350-link" }

Caused by:
    0: Failed to read bootspec file
    1: No such file or directory (os error 2)

Stack backtrace:
   0: anyhow::context::<impl anyhow::Context<T,E> for core::result::Result<T,E>>::context
   1: lanzatool::generation::Generation::from_link
   2: lanzatool::install::Installer::install
   3: lanzatool::cli::Cli::call
   4: lanzatool::main
   5: std::sys_common::backtrace::__rust_begin_short_backtrace
   6: std::rt::lang_start::{{closure}}
   7: std::rt::lang_start_internal
   8: main
   9: __libc_start_call_main
  10: __libc_start_main_alias_1
  11: _start
Malformed generation: Failed to build generation from link: GenerationLink { version: 349, path: "/nix/var/nix/profiles/system-349-link" }

Caused by:
    0: Failed to read bootspec file
    1: No such file or directory (os error 2)

Stack backtrace:
   0: anyhow::context::<impl anyhow::Context<T,E> for core::result::Result<T,E>>::context
   1: lanzatool::generation::Generation::from_link
   2: lanzatool::install::Installer::install
   3: lanzatool::cli::Cli::call
   4: lanzatool::main
   5: std::sys_common::backtrace::__rust_begin_short_backtrace
   6: std::rt::lang_start::{{closure}}
   7: std::rt::lang_start_internal
   8: main
   9: __libc_start_call_main
  10: __libc_start_main_alias_1
  11: _start
'/boot/EFI/nixos/qalw2n6bvbsbvrhz9q9ax82fzxg5xhj0-linux-6.1.2-bzImage.efi' not in use anymore. Removing...
'/boot/EFI/nixos/81r7ygqb3hrzvhqgp0wzm1zmvhj1pzvf-initrd-linux-6.1.2-initrd.efi' not in use anymore. Removing...
'/boot/EFI/nixos/.extra-files' not in use anymore. Removing...
```

nikstur:
Replace `println()` calls with logging at different levels using the `log` crate.

Closes #55 #73

### DIFF
--- a/rust/tool/src/install.rs
+++ b/rust/tool/src/install.rs
@@ -55,6 +55,8 @@ impl Installer {
     }
 
     pub fn install(&mut self) -> Result<()> {
+        log::info!("Installing Lanzaboote to {:?}...", self.esp_paths.esp);
+
         let mut links = self
             .generation_links
             .iter()
@@ -83,6 +85,7 @@ impl Installer {
 
         self.install_systemd_boot()?;
 
+        log::info!("Collecting garbage...");
         // Only collect garbage in these two directories. This way, no files that do not belong to
         // the NixOS installation are deleted. Lanzatool takes full control over the esp/EFI/nixos
         // directory and deletes ALL files that it doesn't know about. Dual- or multiboot setups
@@ -98,6 +101,7 @@ impl Installer {
                     .map_or(false, |n| n.starts_with("nixos-"))
             })?;
 
+        log::info!("Successfully installed Lanzaboote.");
         Ok(())
     }
 
@@ -164,11 +168,15 @@ impl Installer {
                 .with_context(|| format!("Failed to build generation from link: {link:?}"));
 
             // Ignore failing to read a generation so that old malformed generations do not stop
-            // lanzatool from working.
+            // lzbt from working.
             let generation = match generation_result {
                 Ok(generation) => generation,
                 Err(e) => {
-                    println!("Malformed generation: {:?}", e);
+                    log::info!(
+                        "Ignoring generation {} because it's malformed.",
+                        link.version
+                    );
+                    log::debug!("{e:#}");
                     continue;
                 }
             };
@@ -307,7 +315,16 @@ impl Installer {
         ];
 
         for (from, to) in paths {
-            if newer_systemd_boot(from, to)? || !&self.key_pair.verify(to) {
+            let newer_systemd_boot_available = newer_systemd_boot(from, to)?;
+            if newer_systemd_boot_available {
+                log::info!("Updating systemd-boot...")
+            };
+            let systemd_boot_is_signed = &self.key_pair.verify(to);
+            if !systemd_boot_is_signed {
+                log::warn!("systemd-boot is not signed. Replacing it with a signed binary...")
+            };
+
+            if newer_systemd_boot_available || !systemd_boot_is_signed {
                 force_install_signed(&self.key_pair, from, to)
                     .with_context(|| format!("Failed to install systemd-boot binary to: {to:?}"))?;
             }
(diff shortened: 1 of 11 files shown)


rust/tool/src/logging.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    fn log(&self, record: &Record) {
+        if !self.enabled(record.metadata()) {
+            return;
+        };
+        // Do not write the log if it does not come from this crate.
+        if !lzbt_record(record) {
+            return;
+        }
+        println!("{}", record.args());
```
blitz, line 30: Typically logging goes to stderr to avoid buffering log messages. But at that point, you just re-implement [stderrlog](https://docs.rs/stderrlog/0.5.4/stderrlog/). Why not use that? It also gives you the typical Unix behavior of `-v`, `-vv`, `-vvv`.

rust/tool/src/signature.rs
```diff
@@ -61,10 +59,8 @@ impl KeyPair {
             if std::io::stderr().write_all(&output.stderr).is_err() {
                 return false;
             };
-            println!(
-                "Failed to verify signature using sbverify with args `{:?}`",
-                &args
-            );
+            log::info!("Failed to verify signature of {path:?}.");
```
blitz, line 62: This sounds like `error!` or at least `warn!`? I mean if this is not an error what is? :) In case of errors, it also makes sense to dump all information required for debugging that you have at hand immediately.

blitz: But maybe the log message should be at the call site of `verify`, otherwise legitimate uses of verify might spam the log?

nikstur: Right now this function is called only to check if the systemd-boot binary is signed or not. There, it should not be `error` because we just update the binary if it not signed (e.g. because you just installed Lanzaboote). So you're right it should be at the call site. Maybe a warning makes more sense than an info. But there is really nothing for the user to do, that's why it maybe should be an info?

rust/tool/src/logging.rs
nikstur: Never heard of the stderrlog crate before but it looks good. I'm not too fond of the multiple -vvv flags, since I do not believe that a normal CLI user would be interested in anything besides normal output and debug output. However, since Lanzaboote is not really a normal CLI, as it is only called by a module, we might as well implement all log levels.

nikstur:
Thanks for the review @blitz!

rust/tool/src/install.rs
```diff
@@ -307,7 +315,16 @@ impl Installer {
...
         for (from, to) in paths {
-            if newer_systemd_boot(from, to)? || !&self.key_pair.verify(to) {
+            let newer_systemd_boot_available = newer_systemd_boot(from, to)?;
+            if newer_systemd_boot_available {
+                log::info!("Updating systemd-boot...")
+            };
+            let systemd_boot_is_signed = &self.key_pair.verify(to);
+            if !systemd_boot_is_signed {
+                log::warn!("systemd-boot is not signed. Replacing it with a signed binary...")
+            };
```
nikstur, line 325: I don't love this. Do you have any idea how to make this more concise?

blitz: I think it's good enough. I personally would move it into the `if` statement that does the actual installation, but it doesn't really matter.

blitz (approved):
Looks good!

RaitoBezarius (approved):
LGTM! Thanks!

PR #131, opened 2023-03-14
Properly handle fwupd update capsules, take 2

lilyinstarlight:
This is a mostly verbatim copy of #113 (credit to @dasJ for most all of this) that has been updated to use NixOS/nixpkgs#220555 (which has hit nixos-unstable now). It has been successfully tested with a BIOS update (with secureboot enforcing) on a 12th-gen Framework

Also we probably should update the lockfile in this repo first (or let Renovate bot handle it), since this will probably fail eval without that PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ sign all configurations that should be bootable.
 
 `lzbt` lives in `rust/tool`.
 
-### Stub 
+### Stub
 
 When the Linux kernel and initrd are packed into a UKI, they need an
 UEFI application stub. This role is typically filled by
@@ -88,6 +88,12 @@ signature on the Linux kernel and embedding a cryptographic hash of
 the initrd into the signed UKI.
 
 The stub lives in `rust/stub`.
+
+### Fwupd
+
+When both Lanzaboote and `services.fwupd` are enabled, for
+`fwupd.service` a `preStart` will be added that ensures a signed fwupd
+binary is placed in `/run` that fwupd will use.
 
 ## State of Upstreaming to Nixpkgs
 
--- a/nix/modules/lanzaboote.nix
+++ b/nix/modules/lanzaboote.nix
@@ -63,7 +63,7 @@ in
           cp -r ${cfg.pkiBundle}/* /tmp/pki
           ${sbctlWithPki}/bin/sbctl enroll-keys --yes-this-might-brick-my-machine
         ''}
-  
+
         ${cfg.package}/bin/lzbt install \
           --systemd ${config.systemd.package} \
           --systemd-boot-loader-config ${systemdBootLoaderConfig} \
@@ -73,6 +73,21 @@ in
           ${config.boot.loader.efi.efiSysMountPoint} \
           /nix/var/nix/profiles/system-*-link
       '';
+    };
+
+    systemd.services.fwupd = lib.mkIf config.services.fwupd.enable {
+      # Tell fwupd to load its efi files from /run
+      environment.FWUPD_EFIAPPDIR = "/run/fwupd-efi";
+      # Place the fwupd efi files in /run and sign them
+      preStart = ''
+        mkdir -p /run/fwupd-efi
+        cp ${config.services.fwupd.package.fwupd-efi}/libexec/fwupd/efi/fwupd*.efi /run/fwupd-efi/
+        ${pkgs.sbsigntool}/bin/sbsign --key '${cfg.privateKeyFile}' --cert '${cfg.publicKeyFile}' /run/fwupd-efi/fwupd*.efi
+      '';
+    };
+
+    services.fwupd.uefiCapsuleSettings = lib.mkIf config.services.fwupd.enable {
+      DisableShimForSecureBoot = true;
     };
   };
 }
--- a/flake.lock
+++ b/flake.lock
@@ -181,16 +181,16 @@
     },
     "nixpkgs-test": {
       "locked": {
-        "lastModified": 1671812130,
-        "narHash": "sha256-GALBK+qB9rhnB+lVnxdgtMoXCySXughZZ3+qGO1Ke/k=",
-        "owner": "RaitoBezarius",
+        "lastModified": 1679009563,
+        "narHash": "sha256-jizICiQOqUcYFNHRNNOo69bfyNo36iyuRAHem5z68LQ=",
+        "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "e51bf8cc8e2c75192e930ad83ed272938729e7be",
+        "rev": "371d3778c4f9cee7d5cf014e6ce400d57366570f",
         "type": "github"
       },
       "original": {
-        "owner": "RaitoBezarius",
-        "ref": "simplified-qemu-boot-disks",
+        "owner": "NixOS",
+        "ref": "qemu-boot-disk-using-make-disk-image",
         "repo": "nixpkgs",
         "type": "github"
       }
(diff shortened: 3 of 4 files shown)


blitz:
Thanks for bringing this back from the dead! :)

Can you please run `nix flake update --commit-lock-file`? Otherwise, as you already say, this doesn't evaluate. You can use `nix -L flake check` to see whether everything works. Unfortunately, Hercules CI didn't run on this PR. :(

I've enabled lock file maintenance for Renovate, but it will take some time before the bot comes by again.

README.md
```diff
@@ -89,6 +89,11 @@ the initrd into the signed UKI.
 
 The stub lives in `rust/stub`.
 
+### Fwupd
+
+When both Lanzaboote and `services.fwupd` are enabled, `fwupd.service` will get a `preStart` that
```
SuperSandro2000, line 94:
```suggestion
When both Lanzaboote and `services.fwupd` are enabled, for `fwupd.service` a `preStart` will be added that
```

README.md
```diff
@@ -89,6 +89,11 @@ the initrd into the signed UKI.
...
+ensures a signed fwupd binary in /run that fwupd will use.
```
SuperSandro2000, line 95:
```suggestion
ensures a signed fwupd binary is placed in `/run` that fwupd will use.
```

nix/modules/lanzaboote.nix
```diff
@@ -13,6 +13,10 @@ let
     timeout ${toString timeout}
     console-mode ${config.boot.loader.systemd-boot.consoleMode}
   '';
+
+  # This is the fwupd-efi package. We need to get it this way because a user might override services.fwupd.package,
+  # which may cause pkgs.fwupd-efi to be a different package than what the fwupd package has as dependency.
+  fwupd-efi = builtins.head (builtins.filter (x: (x.pname or "") == "fwupd-efi") config.services.fwupd.package.buildInputs);
```
SuperSandro2000, line 19: Why not upstream that fwupd-efi is added to passthru?

lilyinstarlight: That would be the better idea, honestly. I'll make a PR to nixpkgs later to passthru it

lilyinstarlight:
> Can you please run `nix flake update --commit-lock-file`? Otherwise, as you already say, this doesn't evaluate. You can use `nix -L flake check` to see whether everything works. Unfortunately, Hercules CI didn't run on this PR. :(

It looks like https://github.com/RaitoBezarius/nixpkgs/tree/simplified-qemu-boot-disks needs to be rebased since the checks actually use that branch rather than nixos-unstable-small

nix/modules/lanzaboote.nix
lilyinstarlight: I've added a `TODO` for now to change it once a PR to add `passthru.fwupd-efi` to the `fwupd` package makes it to nixos-unstable. Thank you!

lilyinstarlight: Opened as NixOS/nixpkgs#221155

RaitoBezarius:
Ugh, I need to get this merged. I will do it, apologies for the inconveniences.

nix/modules/lanzaboote.nix
dasJ: Aaand merged

lilyinstarlight: I've pushed an update that will hopefully work once flake inputs are updated. Thank you both!

lilyinstarlight:
NixOS/nixpkgs#221155 has made it to nixos-unstable

@RaitoBezarius What is the best option for using an updated nixpkgs input, given the tests are running from that one branch in your fork? (e.g. should that one branch just be rebased?)

RaitoBezarius:
I will rebase that branch and push for the merge of
https://github.com/NixOS/nixpkgs/pull/207039 so we won't need that branch
anymore.

I will take care of this tomorrow morning CEST.
So we can get this merged soon and unblock everyone on that.

Le mer. 15 mars 2023 à 21:21, Lily Foster ***@***.***> a
écrit :

lilyinstarlight:
I've updated the branch for the `nixpkgs-test` input in the flake to NixOS/nixpkgs#207039 and also pinned the `crane` input to the last good revision because it is currently broken with latest `HEAD` (ref: ipetkov/crane#262)

The flake checks all pass and this otherwise should be good now!

nix/modules/lanzaboote.nix
```diff
@@ -74,5 +74,20 @@ in
           /nix/var/nix/profiles/system-*-link
       '';
     };
+
+    systemd.services.fwupd = lib.mkIf config.services.fwupd.enable {
+      # Tell fwupd to load its efi files from /run
+      environment.FWUPD_EFIAPPDIR = "/run/fwupd-efi";
+      serviceConfig.RuntimeDirectory = "fwupd-efi";
```
lilyinstarlight, line 81: @dasJ, just curious is `serviceConfig.RuntimeDirectory` needed?

This seems to clash because the built-in unit already sets `RuntimeDirectory` and so systemd appends them both separated by a `:`, which fwupd doesn't understand how to process multiple runtime dirs and so assumes `/run/fwupd-efi:/run/motd.d` is literal and makes the motd file in that suspicious-looking directory `/run/fwupd-efi:/run/motd.d/85-fwupd`

lilyinstarlight: Ah I think it was to make sure that directory exists. Since it makes more problems than it solves, I've removed it and added a `mkdir -p /run/fwupd-efi` to the pre-start script

Let me know if my analysis there is wrong @dasJ or if this is appropriate

dasJ: Yeah that works as well but I hate the approach of doing mkdir/chmod/chown in prestart. That's why I did it this way. We could also set `PrivateTmp` and just write to /tmp instead?

lilyinstarlight: I like `/run` better than `/tmp` (since it's supposed to be available for the entire lifetime of the daemon and isn't really temporary), but I'll try to think of options that don't use mkdir if you want

lilyinstarlight: Here's a diff for that latter option, for comparison:

```diff
diff --git a/nix/modules/lanzaboote.nix b/nix/modules/lanzaboote.nix
index 02f8f3f..0514548 100644
--- a/nix/modules/lanzaboote.nix
+++ b/nix/modules/lanzaboote.nix
@@ -78,9 +78,16 @@ in
     systemd.services.fwupd = lib.mkIf config.services.fwupd.enable {
       # Tell fwupd to load its efi files from /run
       environment.FWUPD_EFIAPPDIR = "/run/fwupd-efi";
+    };
+
+    systemd.services.fwupd-efi = lib.mkIf config.services.fwupd.enable {
+      # Exist with the lifetime of the fwupd service
+      boundBy = [ "fwupd.service" ];
+      before = [ "fwupd.service" ];
+      # Create runtime directory for signed EFI app
+      serviceConfig.RuntimeDirectory = "fwupd-efi";
       # Place the fwupd efi files in /run and sign them
-      preStart = ''
-        mkdir -p /run/fwupd-efi
+      script = ''
         cp ${config.services.fwupd.package.fwupd-efi}/libexec/fwupd/efi/fwupd*.efi /run/fwupd-efi/
         ${pkgs.sbsigntool}/bin/sbsign --key '${cfg.privateKeyFile}' --cert '${cfg.publicKeyFile}' /run/fwupd-efi/fwupd*.efi
       '';
```

lilyinstarlight:
I've rebased on latest `master` to get the flake lockfile update in #138 and removed all other flake updates from this PR except for the `nixpkgs-test` input update (since it's pointing to a new branch now)

All flake checks still pass locally for me

RaitoBezarius (approved):
This sounds sane to me; untested yet with our flake check.

PR #164, opened 2023-04-26
tool: improve log message about malformed gens

nikstur:
Tells the user which generations are malformed and how to remove them.

Follow up to #161 to implement the improvements discussed there.


blitz (approved):
Nice! That's really useful. 👍

PR #185, opened 2023-05-23
Cleanup flake

nikstur:
Add a proper description to flake.

Use nativeCheckInputs instead of checkInputs because it is more
semantically correct even if checkInputs works with Crane.


RaitoBezarius (approved):
LGTM, thank you!

PR #194, opened 2023-06-08
linux_loader: use device path builder

phip1611:
To demystify the magic constant for the device path of the LINUX_EFI_INITRD_MEDIA protocol, we can replace it by using the device path builder at the corresponding places.

Although this adds a (negligible) overhead during runtime, I think the enhanced readability is worth it. Less magic constants = better, I think :)

Currently, there is the problem that this step needs to be done twice. Once my MR in uefi is merged (https://github.com/rust-osdev/uefi-rs/pull/849), we can do this once and buffer the device path in a global static.

/ ping @blitz  @nikstur


rust/stub/src/linux_loader.rs
```diff
@@ -8,42 +8,20 @@
...
-// 		},
-// 		LINUX_EFI_INITRD_MEDIA_GUID
-// 	}, {
-// 		EFI_DEV_END_PATH,
-// 		EFI_DEV_END_ENTIRE,
-// 		sizeof(struct efi_generic_dev_path)
-// 	}
-// };
-static mut DEVICE_PATH_PROTOCOL: [u8; 24] = [
-    0x04, 0x03, 0x14, 0x00, 0x27, 0xe4, 0x68, 0x55, 0xfc, 0x68, 0x3d, 0x4f, 0xac, 0x74, 0xca, 0x55,
```
phip1611, line 44: I also verified the hex dump: My code produces the same device path

nikstur:
@phip1611 are you still working on this? I'd love to merge it if you think its ready.

blitz:
@phip1611 It would be really awesome, if you can finish this. If there are any blockers, I can help!

rust/uefi/linux-bootloader/src/linux_loader.rs
```diff
@@ -179,6 +154,21 @@ impl InitrdLoader {
     }
 }
 
+/// Builds the device path for the LINUX_EFI_INITRD_MEDIA protocol.
+/// The protocol is associated with the [`LINUX_EFI_INITRD_MEDIA_GUID`].
+fn build_device_path(vec_buf: &mut Vec<u8>) -> &DevicePath {
```
blitz, line 159: This needs a more specific name. `build_linux_initrd_device_path`?

rust/uefi/linux-bootloader/src/linux_loader.rs
```diff
@@ -179,6 +154,21 @@ impl InitrdLoader {
     }
 }
 
+/// Builds the device path for the LINUX_EFI_INITRD_MEDIA protocol.
+/// The protocol is associated with the [`LINUX_EFI_INITRD_MEDIA_GUID`].
```
blitz, line 158: Can you leave the link to the u-boot commit? That's pretty helpful, if people want to understand what's going on here.

blitz (changes requested):
I left some small comments. Will approve/merge when these addressed.

It's too bad that `install_protocol_interface` takes void pointers...

blitz (changes requested):
Mmh. This fails the tests, because Linux cannot find its initrd. So something must be wrong.

You can try any test:

```console
$ nix -L build .\#checks.x86_64-linux.basic
```

Sorry that the CI is currently not automatically running, because we have some issues... (#213) :( @RaitoBezarius is working on it.

RaitoBezarius:
Soon in https://github.com/NixOS/nixpkgs/pull/262839

phip1611:
Hm, it fails with:

```
nixos-disk-image> [    1.381107] Run /init as init process
nixos-disk-image> [    1.381478] EXT4-fs (vda2): mounted filesystem with ordered data mode. Opts: . Quota mode: disabled.
nixos-disk-image> [    1.431234] random: fast init done
nixos-disk-image> [    2.436070] random: crng init done
nixos-disk-image> [    6.458553] reboot: Restarting system
nixos-disk-image> WARNING: Image format was not specified for 'nixos.raw' and probing guessed raw.
nixos-disk-image>          Automatically detecting the format is dangerous for raw images, write operations on block 0 will be restricted.
nixos-disk-image>          Specify the 'raw' format explicitly to remove the restrictions.
nixos-disk-image> EFI stub: Loaded initrd from LINUX_EFI_INITRD_MEDIA_GUID device path
nixos-disk-image> loading kernel modules...
nixos-disk-image> [    0.183084] Invalid ELF header magic: != ELF
nixos-disk-image> [    0.184950] Invalid ELF header magic: != ELF
```

I assume there is a memory corruption, perhaps as the new solution uses the heap to build the device path.

blitz:
> I assume there is a memory corruption, perhaps as the new solution uses the heap to build the device path.

I think the problem is that we install the `DevicePath` protocol and then free the protocol pointer while it's still installed. We avoid this with the `LoadFile2` protocol by putting the interface struct into a pinned `Box`:

```Rust
pub struct InitrdLoader {
    proto: Pin<Box<LoadFile2Protocol>>,
    // ...
```

If you add a similar member here for the `DevicePath`, we should be good. But I also like the C way of just declaring this as global constant. But I guess that's not as straight-forward in Rust.

> I think we should debug this together in a session, @blitz

Sure. Let's do that.

phip1611:
Oh shoot, of course the memory is freed and we have a use after free. How could I not see this... Typical Monday today.. 🙈

phip1611:
Should be fixed now.

rust/uefi/linux-bootloader/src/linux_loader.rs
```diff
@@ -105,12 +115,12 @@ impl InitrdLoader {
         // Linux finds the right handle by looking for something that
         // implements the device path protocol for the specific device
         // path.
+        init_linux_initrd_device_path();
```
blitz, line 118: nit: This crashes for no good reason if `InitrdLoader` is instantiated multiple times. Given that `get_or_init` is idempotent, I would prefer this to not crash as well.

blitz:
@phip1611 I'll approve/merge when:

1. the checks pass (try `nix -L flake check`) (note to self: #124)
  - at least the style check fails at the moment
2. when you squash the fix commit into your initial commit to maintain easy bisectability. 

I have to say that this is not a clear win over the C style of declaring a static variable. It's very verbose and still has sharp edges. But I guess this way it's at least clear how the device path is constructed.

phip1611:
Done.

phip1611:
Rebased.

blitz:
> Rebased.

Please fix style: `nix -L flake check`

(Still waiting for a resolution to #213, but we're getting there.)

phip1611:
@blitz Despite adding `nix-community.cachix.org`, opening the nix shell still takes ages and everything is build from scratch..

image

phip1611:
Or, does this come from the changed rust files? If so, it feels weird that the shell depends on the project itself..

blitz:
> Or, does this come from the changed rust files? If so, it feels weird that the shell depends on the project itself..

I think we still have lzbt in the shell. This was very useful for development, but we can reconsider that. At least for me this is also not useful anymore.

RaitoBezarius:
I honestly never use `lzbt` in the shell anymore too.

blitz:
I'll remove it later when I'm home. (And also merge this PR.)

phip1611:
@blitz hm, sorry. I actually thought that I fixed the style. Weird.

blitz:
@phip1611 I still can't merge it, because it fails the tests. You can try: `nix -L build .\#checks.x86_64-linux.basic`. In general, please run `nix flake check`.

That being said: Given that there are obviously pretty sharp edges here and we keep breaking this, I'm NAKing this for now. I would like to revisit this once we can statically declare `DevicePath` instances without jumping through hoops.

phip1611:
> because it fails the tests

Weird. Locally, it worked.

phip1611:
Oh fuck, I completely messed up the rebase. I'm very sorry, I apologize for the inconvenience I  have caused.

As this MR is closed now, the pushed changes will probably not land here.

phip1611:
Anyway, I don't quite understand the test infrastructure. Apparently it looks like it succeeds, but the log shows information that look as it should be a failure.. hm. nevermind. It's not to important to have this merged.

Lets work on this again when
a) device paths can be constructed at compile-time
b) there is (better) CI in this repo

phip1611:
> There's now better CI in this repo.

What do you mean? I can't see a github workflow file?

RaitoBezarius:
> > There's now better CI in this repo.
> 
> What do you mean? I can't see a github workflow file and [nix-community/lanzaboote/actions](https://github.com/nix-community/lanzaboote/actions) is empty?

We use Hercules-CI. See https://hercules-ci.com/github/nix-community/lanzaboote

phip1611:
I think it would make sense then to have a "fetch CI" in GitHub that fetches the result from Hercules. Otherwise, it is not very intuitive, right? I'd love to see such an convenient integration. Don't know however if it is a quick fix or takes multiple hours to have something solid.

RaitoBezarius:
This is what happens when a new PR is open! :).
It was disabled so old PRs won't see it for now.

PR #198, opened 2023-06-13
Fix eval when aliases are disabled

SuperSandro2000:
```

       at /nix/store/k7h8k8whqw64rmsg6yhkmmdc82lkczc6-source/flake.nix:66:27:

           65|               # linuxArch is wrong here, it will yield arm64 instead of aarch64.
           66|               config = "${pkgs.hostPlatform.qemuArch}-windows";
             |                           ^
           67|               rustc.config = "${pkgs.hostPlatform.qemuArch}-unknown-uefi";
       Did you mean rustPlatform?
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -63,8 +63,8 @@
             inherit system;
             crossSystem = {
               # linuxArch is wrong here, it will yield arm64 instead of aarch64.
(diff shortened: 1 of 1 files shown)


nikstur (approved):
Thank you!

PR #200, opened 2023-06-17
update QUICK_START.md bootspec

Janik-Haag:
Since this is the default now there should probably be a mention that you can skip this with a somewhat recent system. 

Note: There should probably be a version number stated of the version that made bootspec on by default but I don't know which version this was so :shrug:

### DIFF
--- a/docs/QUICK_START.md
+++ b/docs/QUICK_START.md
@@ -98,27 +98,6 @@ This takes a couple of seconds. When it is done, your Secure Boot keys
 are located in `/etc/secureboot`. `sbctl` sets the permissions of the
 secret key so that only root can read it.
 
-### Switching to bootspec
-
-`lzbt` currently doesn't handle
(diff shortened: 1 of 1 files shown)


blitz:
Thanks for the contribution! Can you remove the whole section instead? I don't think it's relevant to keep this information around.

Janik-Haag:
Okay removed it and renamed the commit message.

Janik-Haag:
While we are at it we should probably also remove the sharp edge here: https://wiki.ssdev.org/wiki/N%C3%BCrnberg
since the corresponding issue got closed: https://github.com/nix-community/lanzaboote/issues/39

blitz (approved):

blitz:
@Janik-Haag It's really great that you fix up the quick start. If you can also remove the sharp edges that are now dull, i would also gladly merge that PR. 👍

Janik-Haag:
No problem, happy to help :D 
Thanks for the great Project ^^

PR #219, opened 2023-08-17
tool: only sync ESP filesystem

tilpner:
I noticed lanzaboote was adding quite a bit of time to my nixos-rebuild time, and tracked it to the `sync()` call. That currently syncs all filesystems, even if it only cares about the ESP filesystem. This commit opens the root directory of the ESP filesystem, passes it to [`syncfs`](https://docs.rs/nix/latest/nix/unistd/fn.syncfs.html) instead, which should result in just syncing the ESP filesystem.

- This is mostly untested, but it passes all flake checks
  - I have no idea how to (automatically) test that this still reliably syncs anything
- #204 probably helps more in reducing sync time than this PR will.

### DIFF
--- a/rust/tool/src/install.rs
+++ b/rust/tool/src/install.rs
@@ -1,12 +1,13 @@
 use std::collections::{BTreeMap, BTreeSet};
-use std::fs;
+use std::fs::{self, File};
+use std::os::fd::AsRawFd;
 use std::os::unix::prelude::PermissionsExt;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::string::ToString;
 
 use anyhow::{anyhow, Context, Result};
-use nix::unistd::sync;
+use nix::unistd::syncfs;
 use tempfile::TempDir;
 
 use crate::esp::{EspGenerationPaths, EspPaths};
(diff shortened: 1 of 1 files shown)


rust/tool/src/install.rs
```diff
@@ -161,7 +162,8 @@ impl Installer {
         // Sync files to persistent storage. This may improve the
         // chance of a consistent boot directory in case the system
         // crashes.
-        sync();
+        let boot = File::open(&self.esp_paths.esp).context("Failed to open ESP root directory.")?;
+        syncfs(boot.as_raw_fd()).context("Failed to sync ESP filesystem.")?;
```
tilpner, line 166: I was going to format these over two lines to match the other `.context(...)?` lines, but rustfmt disagrees.

dasJ:
Just a little pointer, switch-to-configuration.pl implements a variable `NIXOS_NO_SYNC` (as outlined [in the docs](https://nixos.org/manual/nixos/stable/#sec-switching-systems)).
I don't expect this to be added in this PR, just wanted to let everyone know that there may be people who want to set this

Mic92:
Runs on my machine.

RaitoBezarius (approved):
This has waited enough, LGTM.

Issue #226, opened 2023-09-24
Editing kernel's command line doesn't work, even when secure boot is disabled


PaulGrandperrin:
It seems that it's meant to work because when secure boot is enabled, pressing "e" doesn't do anything, but when secure boot is disabled, it shows a prompt to edit the command line.
Unfortunately, it only makes the user think the command line has been edited but doesn't actually do it.

So it seems it intended to work this way already, which is great, but there's probably a bug somewhere.

RaitoBezarius:
I guess this is because we are not passing around any command line that is given to us. But can you show screenshots of what you see when you edit the command line?

alois31:
I changed the behavior back in https://github.com/nix-community/lanzaboote/commit/081714cab9ef445f3a576be8f5e1d42be988b5ba, see the commit message for the rationale. I guess the old behavior could be restored conditional on SecureBoot==0, just like we already do (IIRC) for the hash verification of kernel and initrd.

PaulGrandperrin:
I have no opinion about what should be enforced when secure boot is on, but yeah, when it's off, we should be able to edit it.

What's weird is that right now, when secure boot is on, pressing "e" doesn't do anything, but when it's disabled, the line editor works. Making people think it understood that it's allowed because secure boot is off. Unfortunately, it then ignores the edited line.

I solved my immediate issue by using a NixOS specialisation with "rd.rescue".

But I strongly think editing should work as expected. There are many times when it's useful to edit it at boot time.

PaulGrandperrin:
> I guess this is because we are not passing around any command line that is given to us. But can you show screenshots of what you see when you edit the command line?

I can't take screenshots, but to be clear I'm talking about editing in system-boot/lanzaboot, not in the EFI.

When you can select which NixOS generation to boot from, pressing "e".

RaitoBezarius:
> > I guess this is because we are not passing around any command line that is given to us. But can you show screenshots of what you see when you edit the command line?
> 
> I can't take screenshots, but to be clear I'm talking about editing in system-boot/lanzaboot, not in the EFI.
> 
> When you can select which NixOS generation to boot from, pressing "e".

It's clearer now, I was not certain exactly what it was about, at first. No worries.

RaitoBezarius:
[...]

PR #232, opened 2023-10-02
stub: improve handling of insecure boot

alois31:
First, the handling of Secure Boot being inactive is made more explicit, by relying on the SecureBoot variable instead of falling back too LoadImage when hash verification fails. Then, the command line from the loader is used when Secure Boot is not active, to facilitate some debugging and recovery use cases.

Fixes: https://github.com/nix-community/lanzaboote/issues/226


blitz (changes requested):
@alois31 Thanks for this change. This is really useful!

I've rebased this PR and made some changes to it here: https://github.com/nix-community/lanzaboote/tree/insecure-boot

Can you pull those in?

blitz:
Going to merge it and open a PR for some cleanups.

blitz (approved):

PR #236, opened 2023-10-13
Fix build with `documentation.nixos.includeAllModules = true;`

Ma27:
This includes the options of all modules used in the evaluation, not just the ones from `<nixpkgs/nixos>` in the local manual.

Right now this breaks with

    error: attribute 'loader' missing

    at /nix/store/wf59fvxch3l5s7x0pnpfv7b26q6y010x-source/nix/modules/lanzaboote.nix:26:17:

        25|     configurationLimit = mkOption {
        26|       default = config.boot.loader.systemd-boot.configurationLimit;
          |                 ^
        27|       example = 120;

I'm not sure what's up with `config.boot.loader` (had the exact same issue with `disko`), but using `defaultText` is the common workaround for that.

### DIFF
--- a/nix/modules/lanzaboote.nix
+++ b/nix/modules/lanzaboote.nix
@@ -24,6 +24,7 @@ in
 
     configurationLimit = mkOption {
       default = config.boot.loader.systemd-boot.configurationLimit;
+      defaultText = "config.boot.loader.systemd-boot.configurationLimit";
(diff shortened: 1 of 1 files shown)


nikstur (approved):
Not really sure I have understood the problem but it also doesnt matter much if I do. Thank you!

PR #241, opened 2023-10-19
Pin EDK2 202305

blitz:
EDK2 seems to have broken Secure Boot with their recent release. Pin the older version for now to unblock #216.

See #240 for context.


nikstur (approved):
Lgtm

blitz:
Tests look good!

PR #244, opened 2023-10-24
linux_loader: improve code quality + bug fix

phip1611:
While working on #194, I noticed some small room for improvement.

`*const T` and `Option<&T>` have the [same ABI](https://doc.rust-lang.org/reference/type-layout.html), so we can gain some 
more Rust convenience by using it.

Further, a bug is fixed where INVALID_PARAMETER wasn't returned when the BUFFER_SIZE pointer is NULL.

See UEFI 2.10 13.2.2. EFI_LOAD_FILE2_PROTOCOL.LoadFile() for reference.

### DIFF
--- a/rust/uefi/linux-bootloader/src/linux_loader.rs
+++ b/rust/uefi/linux-bootloader/src/linux_loader.rs
@@ -66,24 +66,20 @@ struct LoadFile2Protocol {
 impl LoadFile2Protocol {
     fn load_file(
         &mut self,
-        _file_path: *const FfiDevicePath,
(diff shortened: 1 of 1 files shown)


phip1611:
From executing `nix -L build .\#checks.x86_64-linux.basic`, it looks like this works.

blitz:
For people looking for the ABI guarantees: https://doc.rust-lang.org/std/option/index.html#representation

phip1611:
Any update?

blitz (approved):

PR #245, opened 2023-10-29
Lightweight Shell

blitz:
@phip1611 mentioned that our shell environment is pretty heavy and can compile for quite some time when entered. This PR drops everything from the shell environment that is not immediately useful for compile the Rust code or style checking.

While I was here, I also applied `cargo machete` suggestions.


flake.nix
```diff
@@ -259,8 +243,6 @@
               config.packages.stub
               config.packages.tool
             ];
-
-            TEST_SYSTEMD = pkgs.systemd;
```
nikstur, line 263: We should keep this around, otherwise you can't run the lzbt tests from the shell anymore.

nikstur (changes requested):
Otherwise this looks good

RaitoBezarius (approved):
LGTM modulo what @nikstur said.
Some of the tools are useful for us, developers, e.g. OpenSSL, sbctl overidden, etc. But it's fine I guess.

flake.nix
blitz: Good point. `cargo test` in lzbt passes now and I've also left openssl and sbctl.

blitz:
@nikstur Can you check again?

PR #272, opened 2023-12-26
add framework specific instructions

Mic92:


nikstur:
We should eventually factor this out of the quick start into a separate doc.

blitz:
@nikstur Should we also add instructions on how to update the Secure Boot blacklist? We should probably automate this somehow.

nikstur:
Yes we should. I think fwupd can update dbx somehow now.

blitz:
> Yes we should. I think fwupd can update dbx somehow now.

#274

PR #323, opened 2024-03-10
Refactor tests

nikstur:
Refactor the tests into one file per test. Use the `runTest` API consistently for all tests.


nix/tests/lanzaboote/common/efivariables-helper.nix
```diff
@@ -0,0 +1,25 @@
+''
```
blitz, line 1: nit: This could be a plain `.py` file that you read via `builtins.readFile`. This would allow for tooling to operate on it easier (formatting, etc).

blitz (approved):
This a clear improvement over the status quo! LGTM

nix/tests/lanzaboote/common/efivariables-helper.nix
nikstur: I initially started out turning this into a a .py file but then my default linter screams at me that it can't find `machine` etc. That's why I changed it to this again. I'll leave this as something to improve in the future.

PR #328, opened 2024-03-13
Elaborate on entering Setup Mode

justinas:
I tested Lanzaboote yesterday and was confused as there was no obvious option to "enter Setup Mode". With some Google-fu I found out that [erasing the Platform Key](https://wiki.archlinux.org/title/Unified_Extensible_Firmware_Interface/Secure_Boot#Putting_firmware_in_%22Setup_Mode%22) has the same effect.

### DIFF
--- a/docs/QUICK_START.md
+++ b/docs/QUICK_START.md
@@ -256,6 +256,11 @@ On Framework laptops (13th generation or newer) you can enter the setup mode lik
 
 When you are done, press F10 to save and exit.
 
+#### Other systems
+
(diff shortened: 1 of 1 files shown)


blitz:
It really depends on the system what you have to do. At this point, I think the easiest way forward is to collect screenshots or collect some sort of per-vendor instructions. What do you think?

nikstur:
Is there maybe a collection we can link to? We can't be the first to have thought of this.

I looked at the arch wiki, which I think has a pretty good article but no screenshot collection.

blitz:
@nikstur I'll add a link to the Arch wiki in a separate PR. This PR is an improvement regardless. :)

PR #336, opened 2024-04-02
docs: BitLocker recovery key warning

dweee:
Hi all,

Quickly threw it together on GitHub's web UI.
This should close issue #271. I still haven't used Lanzaboote, so I am unsure if OP's mentioning of removing the immutable attribute is required.


blitz (approved):
Thanks for adding this important warning!

dweee:
For info, at least on the only system I have currently set up (Framework 16), I have not needed to remove the immutable attribute on the EFI variables at `/sys/firmware/efi/efivars/*`.

PR #342, opened 2024-05-16
use /run/pki rather than /tmp/pki

MaxHearnden:
This prevents another unprivileged user from creating /tmp/pki and sbctl enrolling those keys instead

I would be open to using another directory, however at the moment key enrollment is not secure.


nikstur:
Key enrollment like this isn't secure at all and you most likely shouldn't do it. This hook really is only useful for testing (in a VM!) right now.

nikstur:
Closing this as this https://github.com/nix-community/lanzaboote/pull/229 contains the proper long term fix.

PR #355, opened 2024-06-09
Remove an IFD caused by Crane in flake.nix

korfuri:
Before this commit, we use cranelib.cleanCargoSource[1] to filter files in the ./rust/uefi directory. This prevents some rebuilds when files that do not contribute to the build are changed (e.g. a README). However, this causes lanzaboote's flake.nix to contain an IFD[2]. This seems to be unavoidable when using cleanCargoSource[3].

Why this is useful anyway:

Since `nix flake check` must check all outputs of the given flake, it's not possible to split a flake check across multiple builder machines. This means that `nix flake check` without `--no-build` can only work if the builder machine has support to build for all architectures present in that flake's `nixosConfigurations` output. This is achievable with binfmt emulation, but excruciatingly slow, especially for use-cases like "run nix flake check in CI". In practice this means that a flake containing nixosConfiguration outputs with different `system` attributes can generally only be `nix flake check`'d with `--no-build`. See [4] for a long discussion on this.

Unfortunately, this affects flakes using Lanzaboote transitively. So any flake containing Lanzaboote-enabled nixosConfigurations (like mine[5]) are not be able to use `nix flake check --no-build`.

This commit corrects this by removing the IFD. This can be tested with:

$ nix store gc   # to ensure IFD paths are not cached
$ nix flake check --no-build github:nix-community/lanzaboote  # fails
$ nix store gc
$ nix flake check --no-build github:korfuri/lanzaboote  # succeeds

[1] https://crane.dev/API.html#cranelibfiltercargosources
[2] https://nix.dev/manual/nix/2.22/language/import-from-derivation
[3] https://github.com/ipetkov/crane/issues/612
[4] https://github.com/NixOS/nix/issues/4265
[5] https://gitlab.com/korfuri/infra/-/blob/77635232eb65e877d8b79eb98f6264381231c20c/flake.nix

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -142,7 +142,7 @@
 
           stubCrane = buildRustApp {
             pname = "lanzaboote-stub";
-            src = craneLib.cleanCargoSource ./rust/uefi;
+            src = ./rust/uefi;
             target = rustTarget;
             doCheck = false;
           };


korfuri:
Just to clarify, this PR is meant for discussion: I realize this is a trade-off and the pros and cons are not obvious to me. But I'd really like a way out of this IFD so my CI can pass again and I can keep booting my machines securely :)

nikstur:
Am I understanding correctly that this problem exists regardless of the filter function we use?

Completely dropping the filter function sadly isn't really feasible for development purposes.

korfuri:
That is correct, the filter function is not relevant here, the issue lies with how `cleanCargoSource` works, and more broadly, with how Crane operates: it reads a whole bunch of files to generate Nix expressions to decide what to build, which ultimately means it uses an IFD. There's more details in https://github.com/ipetkov/crane/issues/612 but it seems that Crane just needs to live with this load-bearing IFD to do what it does.

I'm wondering, would there be a way to use a filtered source during development, and an unfiltered source in the "published" source? Maybe through some specific branch? This may actually end up being my hacky solution: I can depend on my fork of lanzaboote with this simple change applied, and just sync with upstream regularly.

ipetkov:
This should be fixed with https://github.com/ipetkov/crane/pull/641. As long as you pull in at least https://github.com/ipetkov/crane/commit/a3f0c63eed74a516298932b9b1627dd80b9c3892 (or later) and avoid nesting both `craneLib.cleanCargoSource` and `craneLib.path` (which it doesn't look like you do here) the IFD in the original post should not trigger

korfuri:
Thanks to ipetkov's work in https://github.com/ipetkov/crane/issues/612 this is not needed anymore. https://github.com/nix-community/lanzaboote/pull/360 is another PR that simply bumps the crane input for Lanzaboote and removes the IFD in flake checks. Closing this one.

PR #380, opened 2024-07-22
ci-test

Mic92:


blitz:
@Mic92 Can this be closed?

nikstur:
To give some context: we were debugging why the CI is flaky. More concretely why we see intermittent errors when creating the disk image via make-disk-image.

PR #391, opened 2024-10-01
Minor coding style fix: lib.getExe instead of ${bin}/bin/bin

kuflierl:
This is a minor change in coding style that removes path references replacing them with `lib.getExe`.
This makes executable references less dependent on binary names.

Major advantage: the `cfg.package` option doesn't require the binary name to be "lzbt" anymore.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -37,7 +37,7 @@
         # Derive the output overlay automatically from all packages that we define.
         inputs.flake-parts.flakeModules.easyOverlay
 
-        # Formatting and quality checks. 
+        # Formatting and quality checks.
       ] ++ (if inputs.pre-commit-hooks-nix ? flakeModule then [ inputs.pre-commit-hooks-nix.flakeModule ] else [ ]);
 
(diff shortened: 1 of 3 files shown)


kuflierl:
@ofborg test

blitz:
Looks good. Thanks for this cleanup! One minor thing: can you fix your commit names to match the style of the other commits in the repository? Will merge afterwards!

kuflierl:
> Looks good. Thanks for this cleanup! One minor thing: can you fix your commit names to match the style of the other commits in the repository? Will merge afterwards!

@blitz I believe that everything could be squashed into a single commit. (I was a bit too verbose while making the commits). Would it be fine if I squashed everything into a single commit?

RaitoBezarius (approved):
LGTM modulo commit messages.

kuflierl:
@blitz is the commit name alright?

blitz (approved):
Thank you for taking the time to do this!

PR #398, opened 2024-10-14
linux-bootloader: Clean up remaining uses of RuntimeServices

nicholasbishop:
Missed a few spots in the previous PR.


RaitoBezarius (approved):
Ran the regression test locally.

PR #399, opened 2024-10-15
Drop most uses of `BootServices`

nicholasbishop:
Continuing the migration from previous PRs, clean up most uses of `BootServices` and `boot_services()`. In a couple place it's still needed for the uefi-0.31 API; those can be cleaned up when updating to uefi-0.32.


blitz:
Thank you @nicholasbishop! This refactoring is most welcome!

I'm currently checking what happened to the CI.

blitz (approved):
I shepherded the CI a bit and everything passes.

PS. I really like the fine-grained commits. :heart:

PR #400, opened 2024-10-15
Drop most uses of `SystemTable`

nicholasbishop:
Continuing to smooth the way for updating to uefi-0.32.


blitz (approved):
Thank you for putting so much work in to get the uefi crate updated. It's really appreciated!

PR #422, opened 2025-01-02
Fix activationScript enrollKeys option

itzwam:
this PR fixes the following error when enrollKeys is enabled

```
/nix/store/hk00l6r59x0xbafvd94y9zqp9y3ay09q-bootinstall: line 2: mkdir: command not found
/nix/store/hk00l6r59x0xbafvd94y9zqp9y3ay09q-bootinstall: line 3: cp: command not found
open /tmp/pki/GUID: no such file or directory
Failed to install generation.: No such file or directory (os error 2)
Failed to install bootloader
warning: error(s) occurred while switching to the new configuration
```


Mic92 (approved):

blitz:
So weird that these are not in PATH.

blitz (approved):

PR #440, opened 2025-03-06
Remove Fat Stub

blitz:
People who want the fat stub experience are currently better served with the upstream sd-stub. Remove it until interest in developing a Rust fat stub revives. 

See previous issues with the fat stub here: #420

@nikstur @RaitoBezarius Is there more to remove?


nikstur:
We can remove the whole `uki.nix` module and all associated tests. No need to test the sd-stub here as we don't integrate it with lanzaboote. Nixpkgs has upstream tests for the sd-stub.

nikstur (approved):

PR #443, opened 2025-03-11
Add keepCurrentBootedConfiguration option

jbgi:
to always keep booted system configuration in boot menu.

it is mostly a prof of concept (this is my first time writing some rust code) so probably need some work.

This is a fix/workaround for https://github.com/NixOS/nixpkgs/issues/24158 / https://github.com/NixOS/nix/issues/1095 (only when using lanzaboote of course).

### DIFF
--- a/nix/modules/lanzaboote.nix
+++ b/nix/modules/lanzaboote.nix
@@ -60,6 +60,8 @@ in
       '';
     };
 
+    keepCurrentBootedConfiguration = lib.mkEnableOption "keeping the current booted configuration (labeled 'SAFE' in boot menu)";
+
     pkiBundle = lib.mkOption {
       type = lib.types.nullOr lib.types.externalPath;
       description = "PKI bundle containing db, PK, KEK";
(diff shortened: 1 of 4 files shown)


Majiir:
Why should this be a feature of Lanzaboote or any other boot loader? Couldn't this be a feature of `nix-env --delete-generations`/`nix profile wipe-history` instead? That way, garbage collection is handled robustly (relying on gcroots to profile generations, not just `/run/booted-system` which changes on next boot) and boot loaders don't have to be concerned with how generations are managed.

jbgi:
Probably a solution built in NixOS would be ideal, following comment from a lix developer on this matter: https://git.lix.systems/lix-project/lix/issues/723#issuecomment-9321

nlewo:
I also think this should not be done by lanzaboote, especially since this could lead to non bootable entry. For instance:

1. I boot on `/nix/var/nix/profiles/generation-1`
2. I switch to `/nix/var/nix/profiles/generation-2`
3. I reboot: `/run/booted-system` is no longer a link to `/nix/var/nix/profiles/generation-1`
4. I garbage collect `/nix/var/nix/profiles/generation-1` 
5. I reboot: it fails becasue `/nix/var/nix/profiles/generation-1` doesn't exist anymore

With this option, lanzaboote would propose to boot on `/nix/var/nix/profiles/generation-1` while this generation has been garbage collected...

So, i also think preserving the current booted closure should be handle by deployment tools (`nixos-rebuild/nix-env`, comin (author here), ...).

jbgi:
@nlewo The issue of non-bootable entries is not introduced by this patch though. In your scenario, given a `configurationLimit` >= `2`, `Generation-1` ends-up being a non-bootable entry, with or without this patch. This patch would just add another entry (in last position, so never the default), namely `Generation-SAFE` equivalent to `Generation-1`.
Also your last point (`5.`) only happen if you chose a non default entry (`Generation-1` or `Generation-SAFE`), just rebooting would boot Generation-2, right?

PR #450, opened 2025-04-21
Remove rogue dollar sign

SuperSandro2000:
```
$"/boot/EFI/BOOT/BOOTX64.EFI" is not signed. Replacing it with a signed binary... $"/boot/EFI/systemd/systemd-bootx64.efi" is not signed. Replacing it with a signed binary... Collecting garbage...
```


blitz (approved):

SuperSandro2000:
I didn't see which status check failed 😅

PR #453, opened 2025-05-02
support extraInstallCommands like grub/systemd-boot loader modules

itzwam:


itzwam:
Maybe we should add extraFiles as well ?

itzwam:
I will try to implement one

Note that I'm not super familiar with nix OS tests so it will surely needs some adjustments 👍

nikstur:
Closing for the same reason as https://github.com/nix-community/lanzaboote/pull/483#issuecomment-3540939887

PR #456, opened 2025-05-15
feat: enable support for `XBOOTLDR`

### LINKED ISSUE #173: XBOOTLDR support
As far as I can tell, there's no option to tell Lanzaboot to install the bootloader entries to `/boot`, with the bootloader itself residing on `/efi`.

This is necessary for systems that are dual-boot and have to deal with Windows anemic 512MB ESP.

This is something I implemented for systemd-boot in :  https://github.com/NixOS/nixpkgs/pull/226692

sarunint:
Fixes #173

In some systems, especially those pre-installed with Windows, the EFI System Partition (ESP) might not be large enough to contain the kernels and initrds used to boot.

`XBOOTLDR` is a partition specified by [Boot Loader Specification](https://uapi-group.org/specifications/specs/boot_loader_specification/#the-partitions) to accommodate storing additional boot entries.

Currently, Lanzaboote only supports writing kernels, initrds, and EFI stubs to ESP but not `XBOOTLDR`. This PR adds support for it by

- Add an extra argument to `lzbt-systemd` to denote `$BOOT` partition, which is a placeholder for `XBOOTLDR` or ESP, if `XBOOTLDR` doesn't exists.
- Retrieve `XBOOTLDR` mount point from `boot.loader.systemd-boot.xbootldrMountPoint` and pass it along to `lzbt-systemd` accordingly.

[This workaround](https://github.com/nix-community/lanzaboote/issues/173#issuecomment-1532386210) mentioned in #173 makes the update image for `fwupd` does not work, since `fwupd` writes the update bundle to `ESP/EFI/nixos`, which is bind to `XBOOTLDR` in the workaround, and thus, the bundle isn't recognised by the UEFI firmware on boot.

### DIFF
--- a/nix/modules/lanzaboote.nix
+++ b/nix/modules/lanzaboote.nix
@@ -28,6 +28,11 @@ let
   ]
   ++ cfg.extraEfiSysMountPoints;
 
+  bootMountPoints = [
+    bootMountPoint
+  ]
+  ++ cfg.extraXbootldrMountPoints;
+  
   mkInstallCommand =
     efiSysMountPoint:
     ''
@@ -257,6 +262,15 @@ in
         List of EFI system partition mount points to install the bootloader to (additionally to boot.loader.efi.efiSysMountPoint).
       '';
       default = [ ];
+    };
+
+    extraXbootldrMountPoints = lib.mkOption {
+      type = lib.types.listOf (lib.types.nullOr lib.types.str);
+      description = ''
+        List of Extended Bootloader (XBOOTLDR) mount points to install the boot entries to, paired with each entries in `extraEfiSysMountPoints` (additionally to boot.loader.systemd-boot.xbootldrMountPoint).
+      '';
+      default = lib.genList (_: null) (lib.length cfg.extraEfiSysMountPoints);
+      defaultText = "[ null ] # repeated to match extraEfiSysMountPoints length";
     };
 
     allowUnsigned = lib.mkEnableOption "" // {
--- a/nix/modules/lanzaboote.nix
+++ b/nix/modules/lanzaboote.nix
@@ -32,9 +32,9 @@ let
     bootMountPoint
   ]
   ++ cfg.extraXbootldrMountPoints;
-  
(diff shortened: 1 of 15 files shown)


sarunint:
After nearly a whole year, I finally have time to make this finally works with appropriate tests.

nix/tests/lanzaboote/common/image.nix
```diff
@@ -75,37 +96,47 @@ in
...
     };
-  };
 
   system.image = {
     id = config.system.name;
     version = config.system.nixos.version;
   };
 
-  system.build.espFiles = espFiles;
-
```
sarunint, line 108: I'm not sure why this is needed in the first place.

nix/modules/lanzaboote.nix
```diff
@@ -189,6 +203,15 @@ in
       default = [ ];
     };
 
+    extraXbootldrMountPoints = lib.mkOption {
+      type = lib.types.listOf (lib.types.nullOr lib.types.str);
+      description = ''
+        List of Extended Bootloader (XBOOTLDR) mount points to install the boot entries to, paired with each entries in `extraEfiSysMountPoints` (additionally to boot.loader.systemd-boot.xbootldrMountPoint).
+      '';
+      default = lib.genList (_: null) (lib.length cfg.extraEfiSysMountPoints);
+      defaultText = "[ null ] # repeated to match extraEfiSysMountPoints length";
```
sarunint, line 283: Not sure if this is the correct way to generate a list of `null`s to a specific length

sarunint:
@RaitoBezarius @nikstur @blitz Can you take a look at this? Sorry to tag you all in this.

danjujan:
I tested this today on my framework 13 notebook. It worked great. Thank you @sarunint for creating this PR.

catuhana:
And plans to merge this?

Tommixe:
I'm using it too since one week, it works perfectly. Thanks @sarunint , I hope it get merged soon.

PR #465, opened 2025-07-04
Npins support

Rexcrazy804:
Enables overriding inputs on non flake systems.


Rexcrazy804:
might have to relocate buildRustApp.nix and document how this can be used.

Rexcrazy804:
for anyone interested in seeing how its used in a npins based nixos configuration at its present state, you may refer [this pr](https://github.com/Rexcrazy804/Zaphkiel/pull/52)

blitz (changes requested):
Hi @Rexcrazy804, thanks for the contribution. Unfortunately, we can't accept it in this state, because it duplicates dependency management and we as maintainers would have to bump dependencies in two places.

So while I'm sympathetic to your usecase, my requirement would be to:

1. Not track dependencies in two places.
2. Still be able to use a bot for dependency bumping.

Also, I think some version of flake-compat can already do what you want: https://github.com/nilla-nix/flake-compat?tab=readme-ov-file#usage

Let me document this.

Rexcrazy804:
I suppose we could remove the npins folder and mandate sources to be passed externally, this is typically what will be done on the user's end and they'll probably `npins import-flake` the `flake.lock` of this repo or add em manually. I'll document the required sources. This should resolve the points you've raised

Rexcrazy804:
I've tried working with that fork of flake-compat a while ago but was unable to succeed with the overrides. I'll attempt to try that again and hopefully assist you with documenting it if that is preferred.

blitz:
Does #469 work for you?

Rexcrazy804:
A bit occupied to test it out now but, the expressions there does look good to me

Rexcrazy804:
closing this pr

PR #469, opened 2025-07-24
docs: add docs for input-overriding without flakes

blitz:
This PR adds documentation how to override the nixpkgs used by Lanzaboote, if you don't want to use flakes. For this we use https://github.com/nilla-nix/flake-compat.

Compared to #465 we avoid having to track dependencies in multiple places.

While I was here I also changed our examples to use `npins` instead of `niv`, because this seems to be more popular these days.


nikstur:
Closed in favour of #504

PR #473, opened 2025-07-30
Delete comment

qm3ster:
No it doesn't

### DIFF
--- a/rust/uefi/stub/src/thin.rs
+++ b/rust/uefi/stub/src/thin.rs
@@ -126,7 +126,6 @@ pub fn boot_linux(handle: Handle, dynamic_initrds: Vec<Vec<u8>>) -> uefi::Result
     // Therefore, it is normal to not verify their hashes against a configuration.
 
     /// Compute the necessary padding based on the provided length
-    /// It returns None if no padding is necessary.
     fn compute_pad4(len: usize) -> Vec<u8> {
         vec![0u8; (4 - (len % 4)) % 4]
     }


blitz:
Looks good to me as well. @qm3ster Can you squash your changes into one commit with a commit message that matches the others? I'll give this a try on my system when I get a chance!

rust/uefi/stub/src/thin.rs
```diff
@@ -125,19 +124,15 @@ pub fn boot_linux(handle: Handle, dynamic_initrds: Vec<Vec<u8>>) -> uefi::Result
...
     // Therefore, it is normal to not verify their hashes against a configuration.
 
-    /// Compute the necessary padding based on the provided length
-    fn compute_pad4(len: usize) -> Vec<u8> {
-        vec![0u8; (4 - (len % 4)) % 4]
-    }
-
-    initrd_data.append(&mut compute_pad4(initrd_data.len()));
+    // Pad to align
+    initrd_data.resize(initrd_data.len().next_multiple_of(4), 0);
```
nikstur: I don't think this is correct. Previously we essentially doubled the size of `initrd_data` by appending a vec of zeros of it's own length. Now, we resize it to it's own length. Which basically only aligns the size to a multiple of four but otherwise is  a no-op.

nikstur (changes requested):

qm3ster:
Sorry for being unresponsive, and thank you

PR #477, opened 2025-08-26
 feat: add boot-counting support to lanzaboote

r-vdp:
See https://systemd.io/AUTOMATIC_BOOT_ASSESSMENT/

Boot counting is controlled with a single option, `boot.lanzaboote.bootCounting.initialTries`, if this option is set to a non-zero value, new boot entries will be created with a counter added, set to the value specified in that option. If the option is set to zero, which is also the default, then no boot-counting counters will be added.
Already existing entries will not be modified, so boot counting only applies to new boot entries.

CC: @RaitoBezarius

### DIFF
--- a/nix/modules/lanzaboote.nix
+++ b/nix/modules/lanzaboote.nix
@@ -141,6 +141,18 @@ in
       '';
     };
 
+    bootCounting = {
+      initialTries = lib.mkOption {
+        type = lib.types.ints.u32;
+        default = 0;
+        description = ''
+          The number of boot counting tries to set for new boot entries.
+          Setting this to zero, disables boot counting.
+          See https://systemd.io/AUTOMATIC_BOOT_ASSESSMENT/
+        '';
+      };
+    };
+
     installCommand = lib.mkOption {
       type = lib.types.str;
       readOnly = true;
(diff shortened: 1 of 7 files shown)


blitz:
Interesting! It will take me some time to get to this. If there is a way to split this into smaller PRs that would help speed things up!

r-vdp:
Yeah I can split it, the reason that I didn't is that the earlier commits don't add much value by themselves besides preparing for the boot counting implementation, so I'm not sure that it would make much sense to merge those separately.

But if it would make reviewing easier for you, then I can definitely create individual PRs, more or less corresponding to the commits in this one.

r-vdp:
@blitz I split out #479 but I cannot set the base of this branch to a branch in my fork, so the commits still show up here (and boot counting needs that PR).
But so you could review that PR first if you wish, before this one.

r-vdp:
With #479 merged (thanks @nikstur!), this is now ready for review, @blitz, @RaitoBezarius, @nikstur (for some reason I don't have permission to actually request reviews on this repo, hence the pings).

r-vdp:
Resolved the merge conflicts.

dasJ (approved):

r-vdp:
This is probably also affected by https://github.com/systemd/systemd/issues/40192

I need to see if we can work around it, or if the only solution is to convince upstream to fix this.

TL;DR is that when you set an explicit default boot entry, then starting from 258.2 systemd-boot will boot that entry regardless of any boot counters.

RaitoBezarius (approved):
LGTM.

RaitoBezarius (approved):
LGTM. Thanks.

nikstur:
Fixed the merge conflict. Good to go now.

r-vdp:
Amazing! Now we just need to fix the systemd behaviour to have boot counting work when setting explicit default entries. This is on my to-do list. See https://github.com/systemd/systemd/issues/31215
As is, boot counting will not yet work as expected with systemd >= 258.3.

r-vdp:
Actually, lanzaboote is not affected by https://github.com/systemd/systemd/issues/31215 because we don't actually set any boot default when switching to an old generation.
I think we should add this to lanzaboote by setting an EFI var with the entry to boot, but probably only after we have a fix merged to systemd to make that actually compatible with boot counting.

kirelagin:
Shouldn’t this take the default from the standard NixOS option (or just use it right away instead of introducing a copy here)?

r-vdp:
> Shouldn’t this take the default from the standard NixOS option (or just use it right away instead of introducing a copy here)?

That option is scoped to systemd-boot, and we disable the upstream systemd-boot module when using lanzaboote, so I think it doesn't make much sense to reuse that option.

PR #478, opened 2025-08-31
Small nix cleanups

r-vdp:
- **chore: avoid rec in options definition**
- **chore: avoid `with lib`**


r-vdp:
@blitz I pulled this out of the boot counting PR, it's not really related, just some small improvements (IMHO).

PR #479, opened 2025-08-31
feat: allow specialisations to set their own sort keys

r-vdp:
This fixes https://github.com/DeterminateSystems/bootspec/issues/147 by bumping the bootspec to the latest version.
Upstream PR: https://github.com/DeterminateSystems/bootspec/pull/207

This means that we now capture the full recursive JSON schema for specialisations, including extensions, while the previous type threw this information away.

This is essential for features like boot counting so that users can set separate sort keys for specialisations, to avoid the boot loader falling back to specialisations when a new generation fails to boot.
Often specialisations are special cases that we don't want to boot automatically, so they can now be sorted below all normal generations so that if a generation fails to boot, the boot loader will fall back onto the previous generation.

CC: @blitz @raitobezarius

### DIFF
--- a/nix/tests/lanzaboote/common/image-helper.nix
+++ b/nix/tests/lanzaboote/common/image-helper.nix
@@ -22,6 +22,6 @@
   os.environ['NIX_DISK_IMAGE'] = tmp_disk_image.name
 
   # Enroll keys via systemd-boot by rebooting
-  machine.start(allow_reboot=True)
-  machine.connected = False
+  ${machine.networking.hostName}.start(allow_reboot=True)
+  ${machine.networking.hostName}.connected = False
 ''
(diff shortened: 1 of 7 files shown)


blitz:
Thanks for all the effort here! I wouldn't vendor the bootspec library though unless there is a clear sign that upstream and us have some unresolvable issues. Let's give the upstream PR a bit of time!

r-vdp:
> Thanks for all the effort here! I wouldn't vendor the bootspec library though unless there is a clear sign that upstream and us have some unresolvable issues. Let's give the upstream PR a bit of time!

Yeah, @RaitoBezarius suggested on matrix to vendor this crate when I discussed the issue with specialisation sort keys with him. He mentioned that there is some other lanzaboote-related work already in the pipeline that would also likely require either vendoring this crate or switching upstreams.

I don't know the details though, so I just followed his suggestion.

r-vdp:
@blitz @RaitoBezarius the upstream bootspec PR was merged and a new release is out, so I updated this PR and I think it's good to go now!

nikstur (approved):
Looks good!

PR #482, opened 2025-09-24
flake: use fenix instead of rust-overlay

### LINKED ISSUE #371: Avoid importing nixpkgs
This article contains the details on why this is bad: <https://zimbatm.com/notes/1000-instances-of-nixpkgs>

I personally ran into an issue while trying to use [numtide/nixpkgs-unfree](https://github.com/numtide/nixpkgs-unfree), which does not allow importing.

andre4ik3:
Closes #371

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## 0.5.0 (unreleased)
 
+## Changed
+
+- Exchanged rust-overlay for fenix. When you override your inputs, you now have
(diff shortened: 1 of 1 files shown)


blitz:
This is really nice! Thank you. Will merge after #494.

nikstur:
~~I'm not sure I get the point of this. We already have only one pkgs invocation. If we use legacPackages we also only have one invocation. What do we gain here?~~

Edit: Ah this is to work around crippled dependency injection in flakes.

PR #494, opened 2025-11-15
treewide: use systemd-repart to build images

nikstur:
This let's me run the entire test-suite via `nix flake check` in 1m 55s
on my laptop. I ran the whole suite by changing a systemd-boot setting
(which re-builds all images).

Now the test suite can be consistently run in its entirety without
indefinitely hanging like it did before. This happened both locally as
well as in CI for us.

Also format with nixfmt-rfc-style to be up-to-date with current standards.

Review only c06adacbd6e61d12546a090044ea207a16df5750 as that is the commit that contains the non-formatting changes.


blitz:
Amazing!!! I'll get to this the coming week!

nikstur:
Made it even faster by building compressed erofs images for the Nix Store and mounting `/` on a tmpfs`. Now down to 1m 55s.

nikstur:
> How hard would it be to upstream this?

Not super hard but it doesn't satisfy all use-cases of make-disk-image. It works well with lzbt because lzbt has been designed to work with plain directories. All the other bootloader installers do not (and thus need a VM to build images). It's not a global solution

However, it's on the list ;)

PR #522, opened 2025-12-15
Fix call to lanzaboote from lon

viperML:
We missed this in #504


nikstur:
Please remove the line that passes `system`. You're passing an already instantiated pkgs which already has a system associated.

On a separate note: it should be hostPlatform, not targetPlatform. targetPlatform is basically only used when you work on compilers.

Issue #520, opened 2025-12-12
Failed to autogenerate keys.


tomwrw:
I'm excited about this release (v1.0.0) as it has the potential to remove a few manual steps from my build process. It seems that the generate-sb-keys.service doesn't run.

I'm building my host using nixos-anywhere from my primary workstation (the target host is next to me). I use impermanence with btrfs. I also use luks encryption for the disk.

I assumed the issue was because at the time nixos-anywhere is building the system, the filesystem is mounted in /mnt, so /var/lib/sbctl... won't exist. But upon reading lanzaboote documentation, it seems the auto enrol/gen service kicks in on first boot, in which case my filesystem should be mounted correctly.

I then thought perhaps it was impermanence, but I have persisted /var/lib/sbctl, so it should be available during boot. The journalctl messages below happen after persisted directories are mounted, so I'm not sure it is impermanence.

Any thoughts would be much appreciated.

```
Dec 12 14:10:51 spectre systemd[1]: prepare-sb-auto-enroll.service: Consumed 9ms CPU time, 13.4M memory peak, 8.8M read from disk.
Dec 12 14:10:51 spectre systemd[1]: Failed to start prepare-sb-auto-enroll.service.
Dec 12 14:10:51 spectre systemd[1]: prepare-sb-auto-enroll.service: Failed with result 'exit-code'.
Dec 12 14:10:51 spectre systemd[1]: prepare-sb-auto-enroll.service: Main process exited, code=exited, status=1/FAILURE
Dec 12 14:10:51 spectre systemd[1]: Starting User Login Management...
Dec 12 14:10:51 spectre systemd[1]: Starting Accounts Service...
Dec 12 14:10:51 spectre nsncd[1641]: Dec 12 14:10:51.717 INFO started, config: Config { ignored_request_types: {}, worker_count: 8, handoff_timeout: 10s }, path: "/var/run/nscd/socket"
Dec 12 14:10:51 spectre systemd[1]: Reached target User and Group Name Lookups.
Dec 12 14:10:51 spectre systemd[1]: Reached target Host and Network Name Lookups.
Dec 12 14:10:51 spectre systemd[1]: Started Name Service Cache Daemon (nsncd).
Dec 12 14:10:51 spectre prepare-sb-auto-enroll-start[1647]: couldn't sync keys: open /var/lib/sbctl/keys/db/db.key: no such file or directory
[...]

PR #523, opened 2025-12-15
fix: Updated conditionPathExists for generate-sb-keys.

tomwrw:
Fixes issue described in #520 whereby generate-sb-keys service won't ever trigger on a host that uses impermanence and has /var/lib/sbctl persisted. Instead of a condition based on pkiBundle, it looks for presence of pkiBundle/keys.

### DIFF
--- a/nix/modules/lanzaboote.nix
+++ b/nix/modules/lanzaboote.nix
@@ -263,7 +263,11 @@ in
       wantedBy = [ "multi-user.target" ];
 
       unitConfig = {
-        ConditionPathExists = "!${cfg.pkiBundle}";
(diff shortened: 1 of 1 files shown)


tomwrw:
> This is the right solution.
> 
> However, you seem to have reformatted the file. Please remove the formatting.

Apologies. Can do. What formatter do you prefer?

tomwrw:
> This is the right solution.
> 
> However, you seem to have reformatted the file. Please remove the formatting.

Should be good now. Switched to nixfmt-rfc-style.

nikstur:
Please squash your commits into one and prefix the message with: `module:`

nikstur (approved):

PR #529, opened 2025-12-22
Modify Secure Boot key enrollment command

eljojo:
Hello, I follow this guide at home and fount out that a command from the wiki was missing.

Updated command to include --firmware-builtin option for enrolling Microsoft keys.

This is based on https://wiki.nixos.org/wiki/Limine

### DIFF
--- a/docs/getting-started/enable-secure-boot.md
+++ b/docs/getting-started/enable-secure-boot.md
@@ -62,9 +62,10 @@ your keys to activate Secure Boot. We include Microsoft keys here to
 avoid boot issues.
 
 ```console
-$ sudo sbctl enroll-keys --microsoft
+$ sudo sbctl enroll-keys --microsoft  --firmware-builtin
 Enrolling keys to EFI variables...
-With vendor keys from microsoft...✓
+With vendor keys from microsoft...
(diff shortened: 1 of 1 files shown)


nikstur:
What's the reason to use this? Do you think this should be the general case? Should everone use `--firmware-builtin`?

eljojo:
@nikstur it seemed like a good idea to me to include the OEM's builtin certificates, they could come with revocation lists that come in handy? that being said, this is a bit of a cargo-culting kind of situation. I'm mostly saw it on that wiki and ran it myself, so I thought it'd be good to add it to this wiki.

it's a bit of a philosophical question whether people should trust their OEM's certificates or not, I think it's a sane default to have, and whoever has opinions about it can remove the command.

nikstur (changes requested):
I looked into this and --firmware-builtin is not something you always want to use. One some laptops (e.g. lenovo) it simply doesn't work.

If you want, you can add a notice to the docs that suggest to use the flag for some notebooks (e.g. frameworks).

eljojo:
thanks for taking a closer look and confirming 👍 good thing we caught this.

PR #541, opened 2026-01-21
Add support for multiple NixOS system profiles

leguteape:
Hey everyone. This PR attempts to  add  support for NixOS system profiles to lanzaboote. I had originally written this patch at around August of 2025 for personal use but the folks over at #135 have encouraged me to create a PR here.

Here's a list of the most important changes:

- Get profile names from `system-profiles/` while also being compatible with non-profiled systems
- Since profile names can also have `-` characters, handle any stray `-`s so versions and profile names can be safely extracted
- Add profile names to the `.efi` entries as well as in boot menus in a fashion consistent with `systemd-boot`
- Update the unit and integration tests to accommodate the presence and absence of profiles

I must add that I didn't know any Rust and sorta hacked my way around when I made these changes, following compiler hints and warnings. So, the quality of the my modifications may not be the best. I would, hence, be welcoming of any suggestions aimed towards improving the quality of this patch.

P.S. Here's an image of what the boot menu looks like after my changes (profile names are inside square brackets):

image

- Entry 1 belongs to the default profile, aka the `system` profile.
- Entry 2 belongs to the `Workstation` profile
- Entry 3 belongs to the `Gamestation` profile

### DIFF
--- a/nix/modules/lanzaboote.nix
+++ b/nix/modules/lanzaboote.nix
@@ -25,6 +25,31 @@ let
   mkInstallCommand =
     efiSysMountPoint:
     ''
+      prof_dir="/nix/var/nix/profiles" profiles=""
+
+      # Add all default system profiles, if any
+      if [ "$(${lib.getExe pkgs.findutils} "$prof_dir" -type l -name 'system-*-link' \
+              | ${lib.getExe' pkgs.coreutils "wc"} -l)" -gt 0 ]; then
+        profiles+="$prof_dir/system-*-link "
+      fi
+
+      # Add all extra system profiles, if any
+      if [ -d "$prof_dir/system-profiles" ] &&
+          [ "$(${lib.getExe pkgs.findutils} \
+                 "$prof_dir/system-profiles" -type l -name '*-*-link' \
+               | ${lib.getExe' pkgs.coreutils "wc"} -l)" -gt 0 ]; then
+        profiles+="$prof_dir/system-profiles/*-*-link"
+      fi
+
+      # Display a clear error message if no usable profile can be found
+      if [ "$profiles" = "" ]; then
+        ${lib.getExe' pkgs.coreutils "printf"} "%s %s %s\n" \
+          "Failed to find usable system profiles. Please make sure that" \
+          "system profiles are present under $prof_dir/system-*-link and/or" \
+          "$prof_dir/system-profiles/<profile-name>-*-link." 1>&2
+        exit 1
+      fi
+
       PATH=${config.systemd.package}/lib/systemd:$PATH
       ${cfg.installCommand} \
     ''
@@ -41,7 +66,7 @@ let
           efiSysMountPoint
         ]
       )
-      + " /nix/var/nix/profiles/system-*-link"
+      + " $profiles"
     );
 
   installHook = pkgs.writeShellScriptBin "lzbt" (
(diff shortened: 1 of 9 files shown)


RaitoBezarius:
@rv32ima can I let you do a first pass? Then ping me and I will do another pass so we can get that promptly merged.

nix/modules/lanzaboote.nix
```diff
@@ -23,11 +23,36 @@ let
   ++ cfg.extraEfiSysMountPoints;
 
   mkInstallCommand = efiSysMountPoint: ''
+    prof_dir="/nix/var/nix/profiles" profiles=""
```
nikstur, line 26: This should definitely not be implemented as a shell script. This needs to be done in the rust code

nikstur (changes requested):

nix/modules/lanzaboote.nix
Eveeifyeve: IA, Rust would be best for this case.

leguteape: Hey. Thanks for your suggestion! I am willing to rewrite this in Rust but I would appreciate some pointers regarding where to implement this functionality. If you could suggest the file/files for storing this code, that would be quite helpful.

lewo-hkcm: @nikstur what do you think about removing the `Installer` `generation_links` argument and generating the list of generations [here](https://github.com/nix-community/lanzaboote/blob/04723d4fd6bde2665fef3b25856aeebbd4013c16/rust/tool/systemd/src/install.rs#L112)?

leguteape:
Hey everyone. I have been busy with academic stuff lately and have not had the opportunity to work on this PR. Since that will continue to be the case in the immediate future, I will just close it as I don't want to give the illusion of progress where there is none. If anyone is interested in continuing this further and finds anything in this PR to be of any help, feel free to use whatever you need.

RaitoBezarius:
Thank you for the ping. This PR still greatly interest me, so I may take up the torch. Thanks and good luck!

eveeifyeve:
I have a question about this PR—hope that’s okay. Will it be compatible with alternative NixOS module systems such as NixBSD or Finix? Specifically, would it allow switching to a BSD/Finix system along with NixOS while using Secure Boot without issues, or is that not currently supported?

RaitoBezarius:
> I have a question about this PR—hope that’s okay. Will it be compatible with alternative NixOS module systems such as NixBSD or Finix? Specifically, would it allow switching to a BSD/Finix system along with NixOS while using Secure Boot without issues, or is that not currently supported?

lanzaboote focuses on systemd. That's already not supported and this PR is not the right place to ask this question.

Spaghetto392:
Apologies if this isn't the place to ask, but is there any potential time frame on a version of this patch being pushed to Lanzaboote?

I've been using an ancient version of comin to at least somewhat skirt around the profiles support being an issue and as of a week ago, it finally stopped working. 

I'm currently trying to weigh my options for the best course of action and so far haven't come up with anything.

Unfortunately, I'm not too knowledgeable about Rust. If I were, I'd likely attempt finishing the patch myself.

nix/modules/lanzaboote.nix
nlewo: @nikstur while I agree with you using a shell script is dirty, this PR solves an issue and the main part of the implementation seems to be ok. Also, this doesn't introduce the usage of a shell script because it is already the case (i know this is not a sufficient reason). 

@nikstur it would be really nice if you could reconsider this PR as "good enough", especially because it fixes a real world issue and could be improved later, without any breaking change.

I have been running this PR on several systems since several weeks without any troubles and it makes my systems working with lanzaboote.

nlewo:
I have been running this PR on my (comin) systems since several weeks: it fixes my issue without introducing any regression.

PR #546, opened 2026-02-07
Implement measuring of kernel/initrd content instead of file path

### LINKED ISSUE #505: Correct TPM Measurements
Or maybe the stub should load the referenced files of `.linuxh` and `.initrdh` into  the measurements of `.initrd` and `.linux`. That way they are compatible with systemd-measure

_Originally posted by @arianvp in https://github.com/nix-community/lanzaboote/issues/493#issuecomment-3536531912_

Tert0:
Loads the kernel and intrd from the file system and measures the file content instead of the UKI section containing the file path.
This matches the measuring behavior of `systemd-stub`  and allows `systemd-measure calculate` to predict PCR 11 values and can be used with `systemd-measure sign` to sign valid PCR 11 values to unseal TPM secrets (like LUKS keys). 

Fixes #505

<details>
<summary>Old POC script for testing (replaced by NixOS test)</summary>

## POC script for testing
```bash
#!/usr/bin/env nix-shell
#! nix-shell -i bash
#! nix-shell -p bintools

# Hardcoded path/config (CHANGE THIS FOR TESTING!)
STUB_PATH="/boot/EFI/Linux/nixos-generation-34-nx2gxcyjcla2gntfmecd44sbakhc2b7jp4lw62do7fcdtdgttcha.efi"
PHASE="enter-initrd:leave-initrd" # should be "" on a "normal" NixOS system as pcrphases are not measured by default
PCR_BANK="sha256"

mkdir -p /tmp/systemd-measure-calculate-poc

for section in .initrd .linux .osrel .cmdline; do
objcopy -O binary --only-section=$section $STUB_PATH /tmp/systemd-measure-calculate-poc/$section
echo "Extracted $section from UKI"
done

LINUX_PATH="/boot$(cat /tmp/systemd-measure-calculate-poc/.linux | sed 's/\\/\//g')"
INITRD_PATH="/boot$(cat /tmp/systemd-measure-calculate-poc/.initrd | sed 's/\\/\//g')"

SYSTEMD_MEASURE="$(nix eval --raw nixpkgs#systemd)/lib/systemd/systemd-measure"

$SYSTEMD_MEASURE calculate \
--linux "$LINUX_PATH" \
--initrd "$INITRD_PATH" \
--osrel /tmp/systemd-measure-calculate-poc/.osrel \
--cmdline /tmp/systemd-measure-calculate-poc/.cmdline \
--phase "$PHASE" \
--bank "$PCR_BANK"

$SYSTEMD_MEASURE status \
--bank "$PCR_BANK"

rm -rf /tmp/systemd-measure-calculate-poc

# The hashes printed by the script should match
```

</details>

### DIFF
--- a/rust/uefi/stub/src/thin.rs
+++ b/rust/uefi/stub/src/thin.rs
@@ -1,41 +1,28 @@
+use crate::common::{boot_linux_unchecked, extract_string, get_cmdline, get_secure_boot_status};
 use alloc::vec::Vec;
+use linux_bootloader::pe_section::pe_section;
+use linux_bootloader::uefi_helpers::PeInMemory;
 use log::{error, warn};
 use sha2::{Digest, Sha256};
-use uefi::{CString16, Result, fs::FileSystem, prelude::*};
-
-use crate::common::{boot_linux_unchecked, get_cmdline, get_secure_boot_status};
-use linux_bootloader::pe_section::{extract_string, pe_section};
-use linux_bootloader::uefi_helpers::booted_image_file;
+use uefi::fs::FileSystem;
+use uefi::{CString16, prelude::*};
 
 type Hash = sha2::digest::Output<Sha256>;
 
-/// The configuration that is embedded at build time.
-///
-/// After this stub is built, lzbt needs to embed configuration into the binary by adding PE
-/// sections. This struct represents that information.
-struct EmbeddedConfiguration {
-    /// The filename of the kernel to be booted. This filename is
-    /// relative to the root of the volume that contains the
-    /// lanzaboote binary.
-    kernel_filename: CString16,
+/// UKI Components that are load from PE sections or from the file system and are used to boot the system
+pub struct UkiComponents {
+    /// Kernel data loaded from file system.
+    pub kernel_data: Vec<u8>,
 
-    /// The cryptographic hash of the kernel.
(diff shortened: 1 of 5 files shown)


blitz:
Nice! I didn't have time to look at this in detail, but things I would love to see:

1. A happy path test in the CI.
1. (Optional) Avoiding reading kernel/initrd multiple times from the filesystem. This might require some larger refactoring so I'm also ok with doing this in a separate PR.
1. In case we can't avoid loading the files multiple times, we should make sure that the hashes of what we load and measure into the TPM actually match. We shouldn't have two code paths where one checks the hash and one doesn't. This is a recipe for sadness. :)

Tert0:
Thanks for the feedback!
I added a NixOS test and refactored the stub a bit to load the files only once at the beginning, but I am not sure if (and how) I should further restructure the stub crate.

rust/uefi/stub/src/thin.rs
blitz: Can you move these checks to `load_from_pe()` above and make it fail when hashes don't match? This makes sure we do absolutely nothing with data that doesn't match its hash, which is the best option from a security perspective, from my view.

blitz:
I like the `UkiComponents` struct. 👍🏻

PR #550, opened 2026-02-10
[WIP]: netboot

sjcobb2022:
I have spent some time revamping the old netboot PR from @RaitoBezarius as I would like to use it for a secure nixos deployment.

There are definitely areas in which this PR could be improved, such as moving some of the build command and runtime into it's own file.

The test suite is not currently complete, however I think that it is close.

I would also like to at some point also provide HTTP boot options (as they are also available on qemu and in the uefi crate). This benefits from having a significantly faster transfer speeds as we are not limited by the smaller window sizes of TFTP. I think following a flow similar to [this](https://github.com/danderson/netboot/blob/main/pixiecore/README.booting.md) would be good, but instead we can bypass iPXE entirely.

### DIFF
--- a/rust/uefi/stub/src/thin.rs
+++ b/rust/uefi/stub/src/thin.rs
@@ -95,9 +95,12 @@ pub fn boot_linux(handle: Handle, dynamic_initrds: Vec<Vec<u8>>) -> uefi::Result
     let kernel_data;
     let mut initrd_data;
 
+    let mut has_fs = false;
+
     {
         let file_system_result = uefi::boot::get_image_file_system(handle);
         if let Ok(file_system_handle) = file_system_result {
+            has_fs = true;
             let mut filesystem = FileSystem::new(file_system_handle);
             (kernel_data, initrd_data) = load_via_fs(
                 &mut filesystem,
@@ -124,20 +127,23 @@ pub fn boot_linux(handle: Handle, dynamic_initrds: Vec<Vec<u8>>) -> uefi::Result
         secure_boot_enabled,
     )?;
 
-    // Correctness: dynamic initrds are supposed to be validated by caller,
-    // i.e. they are system extension images or credentials
-    // that are supposedly measured in TPM2.
-    // Therefore, it is normal to not verify their hashes against a configuration.
+    // Only pad if we have an FS.
+    if has_fs {
+        // Correctness: dynamic initrds are supposed to be validated by caller,
+        // i.e. they are system extension images or credentials
+        // that are supposedly measured in TPM2.
+        // Therefore, it is normal to not verify their hashes against a configuration.
 
-    // Pad to align
-    initrd_data.resize(initrd_data.len().next_multiple_of(4), 0);
-    for mut extra_initrd in dynamic_initrds {
-        // Uncomment for maximal debugging pleasure.
-        // let debug_representation = extra_initrd.as_slice().escape_ascii().collect::<Vec<u8>>();
-        // log::warn!("{:?}", String::from_utf8_lossy(&debug_representation));
-        initrd_data.append(&mut extra_initrd);
-        // Extra initrds ideally should be aligned, but just in case, let's verify this.
+        // Pad to align
         initrd_data.resize(initrd_data.len().next_multiple_of(4), 0);
+        for mut extra_initrd in dynamic_initrds {
+            // Uncomment for maximal debugging pleasure.
+            // let debug_representation = extra_initrd.as_slice().escape_ascii().collect::<Vec<u8>>();
+            // log::warn!("{:?}", String::from_utf8_lossy(&debug_representation));
+            initrd_data.append(&mut extra_initrd);
+            // Extra initrds ideally should be aligned, but just in case, let's verify this.
+            initrd_data.resize(initrd_data.len().next_multiple_of(4), 0);
+        }
     }
 
     boot_linux_unchecked(handle, kernel_data, &cmdline, initrd_data)


sjcobb2022:
x86 failing at:
```
vm-test-run-lanzaboote> machine: Guest shell says: b'[ WARN]: stub/src/thin.rs@077: Initrd hash does not match! Continuing anyway.\r\n'
vm-test-run-lanzaboote> machine # EFI stub: ERROR: Failed to load initrd: 0x8000000000000009
vm-test-run-lanzaboote> machine # EFI stub: ERROR: efi_stub_entry() failed!
```

and aarch failing at 

```
vm-test-run-lanzaboote> machine # [PANIC]: panicked at stub/src/thin.rs:186:10:
vm-test-run-lanzaboote> machine: Guest shell says: b'on...\r\n[ WARN]: stub/src/common.rs@070: Secure Boot is not active!\r\n[PANIC]: panicked at stub/src/thin.rs:186:10:\r\nFailed to que'
vm-test-run-lanzaboote> machine # Failed to query kernel file size via TFTP: Error { status: TFTP_ERROR, data: () }
vm-test-run-lanzaboote> machine: Guest shell says: b'ry kernel file size via TFTP: Error { status: TFTP_ERROR, data: () }\r\n'
```

It's strange to me the error is different, maybe the qemu versions on each have differing impls?

I have had this work locally (`nix run .#checks.x86_64-linux.netboot`) before.

On success we should have:
```
       > machine # [ WARN]: stub/src/thin.rs@078: Initrd hash does not match! Continuing anyway.
       > machine: Guest shell says: b'[ WARN]: stub/src/thin.rs@078: Initrd hash does not match! Continuing anyway.\r\n'
       > machine # EFI stub: Loaded initrd from LINUX_EFI_INITRD_MEDIA_GUID device path
       > machine: Guest shell says: b'EFI stub: Loaded initrd from LINUX_EFI_INITRD_MEDIA_GUID device path\r\n'
```

EDIT: on x86 it seems that the VM is out of resources which is why it is crashing. Still unsure as to why a basic TFTP lookup is not working on aarch64 though.

RaitoBezarius:
In general, I wonder, do you really need to go from PXE to NixOS without using a signed iPXE or something like that? Because that would simplify your code I believe, no?

sjcobb2022:
That is a very good point. Would you prefer if this was not included here then?

RaitoBezarius:
Nah, it's really just an open design question. I'm curious about your usecases that makes you need PXE that way. I think any feature is usually a maintenance churn — this one included —, if there's ways to avoid it and just compose well maintained pieces while having a good operational experience, that's what we should seek I think.

sjcobb2022:
I am doing a secure deployment on university premises, and I thought that this might be a promising way to accomplish this goal. Honestly now I think about it, I got a bit pigeonholed with the idea of using lanzaboote for this. Signing the ipxe.efi should be pretty simple (fingers crossed).

I do think that putting the effort into this might be worth it in the long run however, to make netbooting via lanzaboote slightly more accessible. I personally netboot most of my homelab stuff, so being able to secureboot as well would be pretty cool.

RaitoBezarius:
Right, that's fairly understandable. Obviously, don't get me wrong: having good UX to use lanzaboote to sign ipxe.uefi and so on is probably in-scope and interesting! And even if you use iPXE, remember that lanzaboote assumes being booted from a local disk, so there's a slight change that needs to happen to use the right UEFI protocol there.

Beyond that, that should be pretty smooth.

sjcobb2022:
My apologies for the confusion. I was suggesting we continue with the current custom stub approach rather than providing an interface to build and sign ipxe.efi or similar files. If that level of customization is needed, using sbctl directly might be the better option.

The main constraint I'm working with is that my university project requires diskless operation (we're handling potentially harmful assets that cannot be stored on disk), so assuming disk availability is problematic for this use case.

A potential workflow to consider (though it may be overkill):
 - Create a bootstrap.efi that contains the keys
 - If keys haven't been enrolled yet, enroll them at this stage
 - Otherwise, chain to another EFI binary (need to research the implementation) that's signed with those keys

RaitoBezarius:
> The main constraint I'm working with is that my university project requires diskless operation (we're handling potentially harmful assets that cannot be stored on disk), so assuming disk availability is problematic for this use case.

Why does iPXE requires disk availability? You can serve iPXE and daisy chain the rest of your payload?

sjcobb2022:
It doesn't, which is why I am going to go with just ipxe now I think.

RaitoBezarius:
Right. Is there something you want to achieve in lanzaboote in that new context @sjcobb2022 ? I think that just using iPXE on its own is insufficient, FWIW.

Notably because the custom stub has assumptions of being booted from a local disk last time I tried.

sjcobb2022:
I will give it a shot and get back to you on that if that's alright. I should be able to get some working stuff in 2 weeks or so

sjcobb2022:
Current workflow:

SERVER (pxe server):
- Create bootable USB: USB A (nixos)
- On own device do sbctl create-keys
- Upload keys to USB B.

CLIENT:
- Boot CLIENT with USB A
- Mount USB B
- sbctl enroll-keys on mounted files

SERVER:
- Build signed image with keys
- Serve signed image via dnsmasq or dhcpd

Such that any client only needs to enroll it's keys once, and then can secureboot from then on, as all future images can be signed with the same keys.

I think if we are to modify this current branch we would want something like: 

- Server stores signed images and keys.
- Server serves keys via TFTP (maybe not the most secure)
- Client BIOS entered into secureboot setup mode.
- Client netboots custom lanzaboote image (already signed with keys on server).
- Client fetches keys via TFTP of server. - in efi land (probably only needs public keys to check signature since we assume this device will be constantly netbooted, and not responsible for signing it's own images.)
- Stub checks `RuntimeServices::get_variable(EFI_IMAGE_SECURITY_DATABASE_GUID, PK....)` for matches, and sets or panics accordingly. 
  - This is the equivelant of enrolling the keys (and what sbctl does under the hood)
- Boot into kernel

Then every time we boot, we will do that EFI check and determine if we have the correct keys in the security database.

RaitoBezarius:
> Server serves keys via TFTP (maybe not the most secure)

IME, you will find very hard to do Secure Boot setup enrollment remotely that way, things may have changed but usually IIRC this needs user presence.

The way to do Secure Boot public key enrollment on *servers* is to use BMC APIs in principle.

Because of that, I'm not so sure this is in the responsibility of lanzaboote to do this. But I'm open to it as long as it's well isolated.

> Then every time we boot, we will do that EFI check and determine if we have the correct keys in the security database.

You are aware you cannot change PK once you have set it? Any change in PK will require a complete clear of the security database.

Ideally, the way you should do your workflow is:

- You serve on the first boot of said machine an UEFI update capsule to seed the security database.
- On the subsequent boots, you send the EFI stub directly.

lanzaboote doesn't need about your PK/KEK/db initialization logic, that way.

sjcobb2022:
> You are aware you cannot change PK once you have set it? Any change in PK will require a complete clear of the security database.

Yes good point, I was throwing things into the blue with that one.

> Ideally, the way you should do your workflow is:
> 
>     * You serve on the first boot of said machine an UEFI update capsule to seed the security database.
> 
>     * On the subsequent boots, you send the EFI stub directly.

Yeh this would be ideal, this is kind of how iPXE does it, by sending a DHCP option with option 175 which can be tagged. The issue is re-identifying the device. I think that can be done quite easily with dnsmasq by matching on a tag MAC address.

I think that means this PR is probably no longer necessary, as I personally don't think this is the responsibility of lanzaboote anymore.

RaitoBezarius:
The way I do re-identification is with 2 VLAN and having a stateful logic that moves the server enrolled onto the new VLAN (workload) from the provisioning VLAN.

sjcobb2022:
I don't have control of the VLANs for my project unfortunately, but that shouldn't be a massive deal. dnsmasq very easily matches on the MAC address, or I can just add a nice stateful python script or something as a hook.

I think for me the best plan is:
- Image 1: bootstrap which is super minimal nixos with initrd secrets "baked in", boots and enrolls keys then reboots.
- Image 2: Served secure system, which contains all the necessary services, it's efi is signed with the keys.
- Image 3: Netboot server that serves bootstrap system on first request, and the secure system otherwise.

In theory, if I serve all of these as UKIs, then there is no need for iPXE at all (I think).

Perhaps being able to build purely just a signed image in lanzaboote could be beneficial @RaitoBezarius? As in building the standalone UKI given some certs, and then outputing a UKI. Then we only serve the unsigned UKI of the boostrapping serve `.#nixosConfigurations.bootstrap.x86_64-linux.config.system.build.uki` directly via dhcp, and then use lzbt to build a secure the secure configurations UKI, which is signed by the same keys as specified in the bootstrap system's initrd.secrets.

EDIT:

This means that the secure system can (if I am understanding this right), boot from the assigned UKI, even if it does not have lanzeboot on it.

I guess we could say that this is the responsibility of sbctl as they do have the functionality with `sbctl bundle`, but having first class support could be nice.

EDIT again:

Realising that we could quite easily construct a UKI and specify the secure boot keys locally with boot.uki

sjcobb2022:
Hey @RaitoBezarius.

I managed to get my system setup with the below code for my diskless deployments.

I have it split into 2 nix versions for my pxe server, a bootstrap stage and a regular stage. It would probably benefit from being a bit more dynamic (e.g. serving the uki on the fly instead of requiring a reboot). But it suits my needs for the small deployment that I am doing.

I do think that in the future it might be good to streamline this process perhaps (in lanzaboote), but I do not have the time atm to work on this now so I will close this pr.

```nix
  specialisation.bootstrap = {
    inheritParentConfig = true;
    configuration = {
      environment.etc."tftp-boot/bootstrap.efi".source = "${bootstrap.system.build.uki}/${bootstrap.system.boot.loader.ukiFile}";

      services.dnsmasq.settings.dhcp-boot = lib.mkForce ["bootstrap.efi"];
    };
  };
  ```


```nix
  boot.uki.settings = {
    UKI = {
      Initrd = "${config.system.build.netbootRamdisk}/${config.system.boot.loader.initrdFile}";
      SecureBootPrivateKey = "${inputs.certs}/sbctl/keys/db/db.key";
      SecureBootCertificate = "${inputs.certs}/sbctl/keys/db/db.pem";
    };
  };

  system.build.uki =
    lib.mkForce
    (pkgs.runCommand
      config.system.boot.loader.ukiFile {
        buildInputs = [pkgs.sbsigntool];
      } ''
        mkdir -p $out
        ${pkgs.buildPackages.systemdUkify}/lib/systemd/ukify build \
          --config=${config.boot.uki.configFile} \
          --output="$out/${config.system.boot.loader.ukiFile}"
      '');
```

nikstur:
> The way to do Secure Boot public key enrollment on servers is to use BMC APIs in principle.

If you want to do out of band key enrollment, this is the way. And it works because you (implicitly) trust the BMC (or put differently for this to be secure, you have to trust the BMC).

Generally the other option you have is to implement Trust On First Use (TOFU), where the first image you boot enrolls the keys inside and from then on can only boot images signed with these trusted keys.

Lanzaboote offers TOFU capabilities via auto enrolling. Maybe this is the right solution for your use case.

sjcobb2022:
That would imply that the certs would need to be served in the base image though no?

My bootstrap image essentially does just that. The reason I separated it out is because I then reconnect the target devices network back to the network. To me serving a base-image that includes keys in the initrd seems removes the security of serving this altogether.

Having lanzaboote autoenroll JUST on the bootstrap efi would work, if that is what you meant?

Please correct me if I am wrong or if I have misunderstood what you are trying to say.

We could get around this by using something like sops-nix and encrypting the files, but then we come into the same issue of: how do we get the encryption key now onto the host machine without exposing it to the network on the first enrollment.

RaitoBezarius:
> We could get around this by using something like sops-nix and encrypting the files, but then we come into the same issue of: how do we get the encryption key now onto the host machine without exposing it to the network on the first enrollment.

Proper answer: TPM2 LDevID. But it's hard to get on server you do not fully control.

sjcobb2022:
> Proper answer: TPM2 LDevID. But it's hard to get on server you do not fully control.

Yeh I mean I do have control of the servers, it's the network I don't control.

PR #551, opened 2026-02-11
TPM2 PCR Signatures

Tert0:
Implements signing for precalculated PCR values using `systemd-measure sign` with multiple private keys, PCR banks, and systemd boot phase paths and loads the PCR signatures from the PE section `.pcrsig` into a dynamic initrd at `/.extra/tpm2-pcr-signature.json` (path expected by systemd's tools).

This should enable users to bind their LUKS decryption with TPM2 to PCR 11 without re-enrolling the keys on every update.

~~Depends on #546 and currently contains the commit of #546 for the NixOS test to pass.
(Feel free to close the other PR if you prefer to review/merge everything as one PR)~~

### DIFF
--- a/nix/modules/lanzaboote.nix
+++ b/nix/modules/lanzaboote.nix
@@ -17,10 +17,7 @@ let
 
   configurationLimit = if cfg.configurationLimit == null then 0 else cfg.configurationLimit;
 
-  efiSysMountPoints = [
-    espMountPoint
-  ]
-  ++ cfg.extraEfiSysMountPoints;
+  efiSysMountPoints = [ espMountPoint ] ++ cfg.extraEfiSysMountPoints;
 
   mkInstallCommand =
     efiSysMountPoint:
@@ -37,9 +34,7 @@ let
         ++ lib.optionals (cfg.measuredBoot.enable && pcr 4) [
           "--pcrlock-directory=${cfg.measuredBoot.pcrlockDirectory}"
         ]
-        ++ [
-          efiSysMountPoint
-        ]
+        ++ [ efiSysMountPoint ]
       )
       + " /nix/var/nix/profiles/system-*-link"
     );
@@ -88,6 +83,12 @@ let
     ]
     ++ lib.map (pcr: "--pcr=${toString pcr}") cfg.measuredBoot.pcrs
   );
+
+  pcrSignatureConfigFile =
+    if cfg.measuredBoot.pcrSignatures == [ ] then
+      null
+    else
+      json.generate "lanzaboote-pcr-signing-config.json" cfg.measuredBoot.pcrSignatures;
 in
 {
   imports = [
@@ -140,9 +141,7 @@ in
     };
 
     settings = lib.mkOption {
-      type = lib.types.submodule {
-        freeformType = loaderSettingsFormat.type;
-      };
+      type = lib.types.submodule { freeformType = loaderSettingsFormat.type; };
 
       apply = lib.recursiveUpdate options.boot.lanzaboote.settings.default;
 
@@ -152,9 +151,7 @@ in
         editor = config.boot.loader.systemd-boot.editor;
         default = "nixos-*";
       }
-      // lib.optionalAttrs cfg.autoEnrollKeys.enable {
-        secure-boot-enroll = "force";
-      };
+      // lib.optionalAttrs cfg.autoEnrollKeys.enable { secure-boot-enroll = "force"; };
 
       defaultText = ''
         {
@@ -234,7 +231,10 @@ in
           --systemd-boot-loader-config ${loaderConfigFile} \
           --configuration-limit ${toString configurationLimit} \
           --allow-unsigned ${lib.boolToString cfg.allowUnsigned} \
-          --bootcounting-initial-tries ${toString cfg.bootCounting.initialTries}'';
+          --bootcounting-initial-tries ${toString cfg.bootCounting.initialTries} \
+          ${lib.optionalString (
+            pcrSignatureConfigFile != null
+          ) "--pcr-signature-config ${pcrSignatureConfigFile}"}'';
       defaultText = lib.literalExpression ''
         ''${lib.getExe config.boot.lanzaboote.package} ''${lib.optionalString (config.boot.lanzaboote.logLevel == "debug") "-vv"} install \
           --system ''${config.boot.kernelPackages.stdenv.hostPlatform.system} \
@@ -242,7 +242,10 @@ in
           --systemd-boot-loader-config ''${loaderConfigFile} \
           --configuration-limit ''${toString configurationLimit} \
           --allow-unsigned ''${lib.boolToString config.boot.lanzaboote.allowUnsigned} \
-          --bootcounting-initial-tries ''${toString config.boot.lanzaboote.bootCounting.initialTries}'';
+          --bootcounting-initial-tries ''${toString config.boot.lanzaboote.bootCounting.initialTries} \
+          ''${lib.optionalString (
+            pcrSignatureConfigFile != null
+          ) "--pcr-signature-config ''${pcrSignatureConfigFile}"}'';
     };
 
     extraEfiSysMountPoints = lib.mkOption {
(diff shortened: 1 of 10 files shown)


blitz:
Marking as draft as long as the dependency is not in the tree.

nix/modules/lanzaboote.nix
```diff
@@ -170,15 +233,19 @@ in
...
       defaultText = lib.literalExpression ''
         ''${lib.getExe config.boot.lanzaboote.package} install \
           --system ''${config.boot.kernelPackages.stdenv.hostPlatform.system} \
           --systemd ''${config.systemd.package} \
           --systemd-boot-loader-config ''${loaderConfigFile} \
           --configuration-limit ''${toString configurationLimit} \
           --allow-unsigned ''${lib.boolToString config.boot.lanzaboote.allowUnsigned} \
-          --bootcounting-initial-tries ''${toString config.boot.lanzaboote.bootCounting.initialTries}'';
+          --bootcounting-initial-tries ''${toString config.boot.lanzaboote.bootCounting.initialTries} \
+          ''${lib.optionalString (pcrSignatureConfigFile != null) "--pcr-signature-config ''${pcrSignatureConfigFile}"}'';
```
nikstur: Please stay consistent here.

nix/modules/lanzaboote.nix
```diff
@@ -366,5 +433,22 @@ in
     services.fwupd.uefiCapsuleSettings = lib.mkIf config.services.fwupd.enable {
       DisableShimForSecureBoot = true;
     };
+
+    # Copy stub provided metadata from /.extra into /run/, so that it will survive the initrd stage
+    # Source: https://github.com/systemd/systemd/blob/v259.1/tmpfiles.d/20-systemd-stub.conf.in
+    boot.initrd.systemd.tmpfiles.settings."20-lanzaboote-stub" = {
```
nikstur, line 700: Why would you want that? Can you add that to the comment please?

nix/tests/lanzaboote/systemd-cryptsetup-tpm2-signature.nix
```diff
@@ -0,0 +1,119 @@
...
+          ];
+        };
+      };
+      boot.initrd.systemd.services."systemd-cryptsetup@".before = [
+        "cryptsetup.target"
+        "initrd-switch-root.target"
+      ];
+
+      # Measure boot phases
+      boot.initrd.systemd.storePaths = [ "${config.systemd.package}/lib/systemd/systemd-pcrextend" ];
```
nikstur, line 70: This should be exposed in the lanzabooe module via an option.

nix/modules/lanzaboote.nix
```diff
@@ -153,6 +159,63 @@ in
...
+            };
+            phases = lib.mkOption {
+              type = lib.types.listOf lib.types.str;
+              description = ''
+                Boot phase paths (colon-separated) to sign a PCR policy for.
+                If empty, defaults to default phases of `systemd-measure(1)`.
+
+                Note: By default NixOS does not measure boot phases into PCR 11. See `systemd-pcrphase.service(8)`.
+              '';
+              default = [ ];
```
nikstur: The default should be that the key is only available in the initrd.

nix/modules/lanzaboote.nix
```diff
@@ -153,6 +159,63 @@ in
       };
     };
 
+    pcrSignatures = lib.mkOption {
+      type = lib.types.listOf (
```
nikstur: I don't like that this is a list. Instead, we should use the ability of the pcrsig json to contain multiple policies to include policies for all generations.

rust/tool/shared/src/pe.rs
```diff
@@ -4,20 +4,22 @@ use std::os::unix::fs::MetadataExt;
...
 use std::process::Command;
 
+use crate::pcr_signature::create_pcr_policy_signature_with_systemd_measure;
+use crate::utils::{SecureTempDirExt, file_hash, tmpname};
 use anyhow::{Context, Result};
 use goblin::pe::PE;
 use serde::{Deserialize, Serialize};
 use tempfile::TempDir;
 
-use crate::utils::{SecureTempDirExt, file_hash, tmpname};
```
nikstur, line 12: Please keep these sorted. Internal crates belong below external crates.

rust/tool/shared/src/pcr_signature.rs
```diff
@@ -0,0 +1,122 @@
...
+            args.push(OsString::from("--phase"));
+            args.push(phase.into());
+        }
+
+        for bank in pcr_signature_config_entry.banks {
+            args.push(OsString::from("--bank"));
+            args.push(bank.into());
+        }
+
+        let command = Command::new(format!("{}/lib/systemd/systemd-measure", systemd.display()))
```
nikstur: We should just read this from PATH. See how I did it for pcrlock.

nix/modules/lanzaboote.nix
```diff
@@ -153,6 +159,63 @@ in
       };
     };
 
+    pcrSignatures = lib.mkOption {
```
nikstur: This belongs under the measuredBoot option and needs an enable option.

nix/tests/fixtures/tpm2-pcr-keys/tpm2-pcr-initrd-private-key.pem
nikstur, line 1: I would rather not vendor them and create them in the test.

rust/tool/shared/src/pcr_signature.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub pcrs: Vec<u8>,
+    pub pkfp: String,
+    pub pol: String,
+    pub sig: String,
+}
+
+pub type PcrPolicySignature = HashMap<String, Vec<PcrPolicySignatureEntry>>;
+
+/// Combine multiple PCR policy signatures into one (and remove duplicates)
+pub fn combine_pcr_policy_signatures(
```
nikstur: Shouldn't be pub

```suggestion
fn combine_pcr_policy_signatures(
```

rust/tool/shared/src/pcr_signature.rs
```diff
@@ -0,0 +1,122 @@
...
+
+#[derive(Serialize, Deserialize, PartialEq)]
+pub struct PcrPolicySignatureEntry {
+    pub pcrs: Vec<u8>,
+    pub pkfp: String,
+    pub pol: String,
+    pub sig: String,
+}
+
+pub type PcrPolicySignature = HashMap<String, Vec<PcrPolicySignatureEntry>>;
```
nikstur: Shouldn't be pub

rust/tool/shared/src/pcr_signature.rs
```diff
@@ -0,0 +1,122 @@
...
+                    result_measurement_objects.push(measurement_object);
+                }
+            }
+        }
+    }
+    result
+}
+
+/// Create a PCR policy signature using `systemd-measure sign` from the PE section files and section data files
+pub fn create_pcr_policy_signature_with_systemd_measure(
```
nikstur: This is quite a long name. Better to pick a shorter one.

```suggestion
pub fn create_pcrsig(
```

rust/tool/shared/src/pcr_signature.rs
```diff
@@ -0,0 +1,122 @@
...
+    systemd: &Path,
+    kernel_cmdline_path: &Path,
+    kernel_path: &Path,
+    initrd_path: &Path,
+    os_release_path: &Path,
+    pcr_signature_config_path: &Path,
+) -> Result<Option<Vec<u8>>> {
```
nikstur: This should only return a Result. Either it creates a policy or it doesn't.

rust/tool/shared/src/pe.rs
```diff
@@ -112,8 +119,28 @@ pub fn lanzaboote_image(
     let kernel_path_offs = initrd_path_offs + file_size(&initrd_path_file)?;
     let initrd_hash_offs = kernel_path_offs + file_size(&kernel_path_file)?;
     let kernel_hash_offs = initrd_hash_offs + file_size(&initrd_hash_file)?;
-
-    let sections = vec![
+    let pcr_signature_offs = kernel_hash_offs + file_size(&kernel_hash_file)?;
+
+    let pcr_signature_section =
+        if let Some(pcr_signature_config_path) = &stub_parameters.pcr_signature_config_path {
```
nikstur, line 125: This is an odd construction. If there is a config we should get a pcrsig. If this fails, the whole program should error.

rust/uefi/stub/src/thin.rs
```diff
@@ -90,13 +95,25 @@ fn check_hash(data: &[u8], expected_hash: Hash, name: &str, secure_boot: bool) -
...
 
+    if let Some(pcr_signature) = components.pcr_signature {
+        let cpio = pack_cpio_literal(
+            &pcr_signature,
+            uefi::fs::Path::new(cstr16!("tpm2-pcr-signature.json")),
+            ".extra",
+            555,
+            444,
+        )
+        .expect("Failed to pack cpio from PCR signature data");
```
nikstur: Should the stub really panic here? If yes, please add a comment why this is appropriate.

nikstur (changes requested):
Very nice, thank you! I think this can be eventually nicely combined with pcrlock to add phases support.

nix/modules/lanzaboote.nix
Tert0: I'm sorry, but I'm not sure what you are referring to here.

nix/modules/lanzaboote.nix
Tert0: I'm not sure how to properly implement the module without a list. The lanzaboote tool already merges all signatures created based on this config into a single JSON PCR signature file (per generation).

How would you implement the module without a list to support multiple keys for different boot stages? 
Example:
```nix
{
  pcrSignatures = [
    {
      privateKeyFile = "/etc/systemd/private-key-initrd.pem";
      phases = [ "enter-initrd" ];
    }
    {
      privateKeyFile = "/etc/systemd/private-key.pem";
      phases = [
        "enter-initrd"
        "enter-initrd:leave-initrd"
        "enter-initrd:leave-initrd:sysinit"
        "enter-initrd:leave-initrd:sysinit:ready"
      ];
      banks = [ "sha512" ];
    }
  ];
}

```

nix/modules/lanzaboote.nix
nikstur: The defaultText and default option are not consistently formatted here.

nix/modules/lanzaboote.nix
nikstur: > How would you implement the module without a list to support multiple keys for different boot stages?

I wouldn't. I would just support a single key that you can use to bind your LUKS encryption to.

Tert0: While I'd like to have support for multiple keys to bind different LUKS devices to different boot stages, if this is blocking, I'd refactor this PR to remove the list.

PR #555, opened 2026-02-27
rust: various cosmetic improvements to dependency management

phip1611:
- use resolver v3 (default for 2024 edition)
- use workspace dependencies, remove duplicates
- sort deps alphabetically
- group dependencies
  - workspace
  - normal
  - path

### DIFF
--- a/rust/tool/Cargo.toml
+++ b/rust/tool/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-resolver = "2"
+resolver = "3"
 
 members = [
     "shared",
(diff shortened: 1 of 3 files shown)


rust/uefi/Cargo.lock
```diff
@@ -118,7 +118,6 @@ name = "linux-bootloader"
 version = "1.0.0"
 dependencies = [
  "bitflags",
- "embedded-io",
```
phip1611: hints for reviewers: dependency was unused

PR #558, opened 2026-03-05
[Docs] More familiar flakes example

rrueger:
...in particular the example now resembles the structure created by `nixos-generate-config --flake` more closely.

Indeed, there, the structure is

```
$ sudo nixos-generate-config --flake
$ ls -l /etc
flake.nix
configuration.nix
hardware-configuration.nix
```

and appears to be very common outsie the `nixos-generate-config` context.


rrueger:
Sorry for (embarrassing) typos. Thanks for spotting! Fixed and pushed.

nikstur:
@rrueger thank you for your contribution! 

I don't see how this improves the docs. IMO it doesn't make it more readable, it just adds steps that feel like noise.

PR #563, opened 2026-03-26
lzbt: with boot assessment, preserve at least one known-good boot entry when applying configuration limit

r-vdp:
With boot assessment enabled, applying the configuration limit can evict
all known-good entries from the ESP, leaving only unassessed entries
with boot counting suffixes. If all of those fail assessment, the system
can become unbootable.

Scan the ESP for stubs without a boot counting suffix (+N) before
selecting which generations to keep. If none of the latest N generations
(N = configuration_limit) are known-good, but known-good entries do exist,
then also include the most recent known-good generation as a fallback.
This means that we may exceed the configured limit by one entry.

### DIFF
--- a/rust/tool/systemd/src/install.rs
+++ b/rust/tool/systemd/src/install.rs
@@ -120,16 +120,37 @@ impl<S: Signer> Installer<S> {
 
         // A configuration limit of 0 means there is no limit.
         if self.configuration_limit > 0 {
-            // Only install the number of generations configured. Reverse the list to only take the
-            // latest generations and then, after taking them, reverse the list again so that the
-            // generations are installed from oldest to newest, i.e. from smallest to largest
-            // generation version.
-            links = links
-                .into_iter()
+            // Only install the number of generations configured.
+            let mut versions_to_keep: BTreeSet<u64> = links
+                .iter()
                 .rev()
                 .take(self.configuration_limit)
-                .rev()
-                .collect()
+                .map(|l| l.version)
+                .collect();
+
+            // When boot counting is enabled, ensure at least one known-good entry
+            // is preserved to avoid an ESP with only unassessed boot entries.
+            // We scan the ESP here because the known-good status (absence of boot
+            // counting suffix) is on-disk state set by systemd-boot, and we need
+            // this information before deciding which generations to keep.
+            if self.bootcounting_initial_tries > 0 {
+                let known_good = self.find_known_good_generation_versions()?;
+                if !known_good.is_empty()
+                    && !versions_to_keep.iter().any(|v| known_good.contains(v))
+                {
+                    if let Some(fallback) =
+                        links.iter().rev().find(|l| known_good.contains(&l.version))
+                    {
+                        log::info!(
+                            "Preserving known-good generation {} as a bootable fallback.",
+                            fallback.version
+                        );
+                        versions_to_keep.insert(fallback.version);
+                    }
+                }
+            }
+
+            links.retain(|l| versions_to_keep.contains(&l.version));
         };
         self.install_generations_from_links(&links)?;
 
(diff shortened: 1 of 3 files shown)


Stebalien:
I'd consider keeping the currently booted system (`/run/booted-system`, not `/run/current-system`) instead of relying on boot counting. Unless configured correctly, boot counting doesn't guarantee anything beyond "it booted with no failed services". Importantly, it doesn't guarantee internet access, a working graphical session, a stable system, etc. The [docs](https://systemd.io/AUTOMATIC_BOOT_ASSESSMENT/) suggest that graphical environments should implement some kind of "wait and see" boot assessment logic, but I don't think anyone has actually implemented that yet.

I'm imagining an scenario where:

1. The user boots a kernel with broken WiFi drivers. It "boots" but the user is unable to actually use the internet.
2. The user spends a few minutes debugging. The boot is marked as "good" (automatically) because it's "usable" (the user is actively using it).
3. The user gives up and reboots into a previous generation to get back working internet while they try to fix their problem (e.g., because they need to download packages, apply patches, etc.).
4. The user messes with their config, switching generations multiple times, eventually wiping the currently booted (working) boot-loader entry.
5. The user finally reboots again and has no boot-loader entries for a kernel with working WiFi drivers.

On the other hand, if we keep the currently booted system, the user will always be able to return to the current point in their debugging process (there is a path from the booted system to the current attempt to activate a new system).

r-vdp:
I think we should assume that if someone is using boot counting, that they configure it correctly for their needs, instead of having workarounds based on second-guessing how people use their systems.

The currently booted system might not actually be a "known good" one even if the user can interactively run nixos-rebuild from it, since `boot-complete.target` might not have been reached.

So I think that the current implementation is more configurable and reliable than assuming that the currenly booted system is always a safe option.

Stebalien:
I assumed boot counting would end up being enabled by default (in a naive configuration) to prevent users from losing all bootable generations. Given that that won't be the case...

How about, if boot counting is disabled, keep the currently booted generation? Without this, calling `nixos-rebuild switch` 7 times before rebooting can easily end up in an unbootable system. I'm happy to take a stab at this at some point in a separate PR if this is acceptable.

> The currently booted system might not actually be a "known good" one even if the user can interactively run nixos-rebuild from it, since boot-complete.target might not have been reached.

You make a good point here but I think there are actually two definitions of "good":

1. The generation marked as "good" by boot counting likely starts all the services correctly; this is what you'd rollback to if you just needed to restore everything to a "working" state again. But that "working" state might not have a fully working nix (e.g., switching to a new generation might require connecting to a build server on a VPN, etc.).
2. The currently booted generation is known to have a working `nix`. More specifically, we _know_ it's possible to successfully call `nixos-rebuild switch` (and similar) from this generation.

I feel less strongly about this but, if boot counting is enabled, I'd actually suggest keeping _both_ the last generation marked as good _and_ the currently booted generation (if they differ).

r-vdp:
I'm not sure it makes much sense to enable boot counting by default, because it doesn't do much without an oob and unattended mechanism to reboot the machine, which lanzaboote doesn't provide.

Besides that, I think that systemd's boot counting has an unambiguous and configurable definition of "good", namely that a specific target was reached (with the caveat that if that depends on external factors, like network resources, things become non-deterministic).
I think that we should just stick to this definition and not make things more confusing by having different kinds of "good".
You configure what "good" means by having `boot-complete.target` depend on whatever you consider essential, which means that it can be whatever makes sense in your environment (you can probably even encode what you want here).

Your definition of "good" basically means "the system made it to stage 2", which is most of the time not a good estimator of whether the generation is "good", and is also not at all configurable.

Stebalien:
I'm not suggesting that boot counting be enabled by default. I don't think it should be as it's not very useful without additional configuration.

I'm suggesting that lanzaboote should at least _try_ to keep one known "bootable" boot entry for the vast majority of users who won't have boot counting enabled.

> Your definition of "good" basically means "the system made it to stage 2", which is most of the time not a good estimator of whether the generation is "good", and is also not at all configurable.

My definition is: the system made it to the point where the user was able to call `nixos-rebuild switch`. I guess "recoverable without a live CD" would be a better definition than "good".

Eveeifyeve:
> I'm not suggesting that boot counting be enabled by default. I don't think it should be as it's not very useful without additional configuration.
> 
> I'm suggesting that lanzaboote should at least _try_ to keep one known "bootable" boot entry for the vast majority of users who won't have boot counting enabled.
> 
> > Your definition of "good" basically means "the system made it to stage 2", which is most of the time not a good estimator of whether the generation is "good", and is also not at all configurable.
> 
> My definition is: the system made it to the point where the user was able to call `nixos-rebuild switch`. I guess "recoverable without a live CD" would be a better definition than "good".

I 100% agree all with this, Because I had an issue with a gpu driver (amdvlk -> radv) not being able to see my screen I had to go into a live CD to fix it, but I could see something like a wifi driver be more serious not being able to switch your config would be devastating.

So I would perfer to see it do checks if a driver has failed or lets say a driver that actually works (eg. linking back to my incident, checking to able to see my screen for example).

r-vdp:
I would suggest to implement that in another PR then, potentially gated by an option.

Let's keep this PR focused on systemd-based boot assessment.

jbgi:
> I would suggest to implement that in another PR then, potentially gated by an option.

btw, that other PR (keep current booted config) has been oppened for a while now: https://github.com/nix-community/lanzaboote/pull/443

r-vdp:
@nikstur @blitz any chance to get a review and/or merge on this?

blitz:
> @nikstur @blitz any chance to get a review and/or merge on this?

Currently occupied. I can get back to things next week. Maybe @nikstur has time before that?

PR #564, opened 2026-03-28
Measured Boot

nikstur:
Adds support for Measured Boot via systemd-pcrlock. Uses PCR4 to measure boot artifacts.

Includes support for automatically enrolling and unattended system into a measured boot state.

 Support should still be considered experimental because pcrlock is still experimental and I'm lacking experience with it.

### DIFF
--- a/nix/modules/lanzaboote.nix
+++ b/nix/modules/lanzaboote.nix
@@ -22,19 +22,62 @@ let
   ]
   ++ cfg.extraEfiSysMountPoints;
 
-  mkInstallCommand = efiSysMountPoint: ''
-    ${cfg.installCommand} \
-      --public-key ${cfg.publicKeyFile} \
-      --private-key ${cfg.privateKeyFile} \
-      ${efiSysMountPoint} \
-      /nix/var/nix/profiles/system-*-link
-  '';
+  mkInstallCommand =
+    efiSysMountPoint:
+    ''
+      PATH=${config.systemd.package}/lib/systemd:$PATH
+      ${cfg.installCommand} \
+    ''
+    + (
+      lib.escapeShellArgs (
+        [
+          "--public-key=${toString cfg.publicKeyFile}"
+          "--private-key=${toString cfg.privateKeyFile}"
+        ]
+        ++ lib.optionals (cfg.measuredBoot.enable && pcr 4) [
+          "--pcrlock-directory=${cfg.measuredBoot.pcrlockDirectory}"
+        ]
+        ++ [
+          efiSysMountPoint
+        ]
+      )
+      + " /nix/var/nix/profiles/system-*-link"
+    );
 
   format = pkgs.formats.yaml { };
   sbctlConfigFile = format.generate "sbctl.conf" {
     keydir = "${cfg.pkiBundle}/keys";
     guid = "${cfg.pkiBundle}/GUID";
   };
+
+  json = pkgs.formats.json { };
+
+  pcr = n: lib.elem n cfg.measuredBoot.pcrs;
+
+  staticMeasurements = pkgs.runCommand "pcrlock.d" { preferLocalBuild = true; } ''
+    mkdir -p $out
+
+    for f in ${toString cfg.measuredBoot.upstreamStaticMeasurements}; do
+      mkdir -p $(dirname $out/$f)
+      ln -sf ${config.systemd.package}/lib/pcrlock.d/$f $out/$f
+    done
+
+    ${lib.concatLines (
+      lib.mapAttrsToList (n: v: "ln -s ${v.source} $out/${n}.pcrlock") cfg.measuredBoot.staticMeasurements
+    )}
+  '';
+
+  makePolicyCommand = lib.escapeShellArgs (
+    [
+      "${config.systemd.package}/lib/systemd/systemd-pcrlock"
+      "make-policy"
+      "--components=${staticMeasurements}"
+      "--components=${cfg.measuredBoot.pcrlockDirectory}"
+      "--policy=${cfg.measuredBoot.pcrlockPolicy}"
+      "--location=770"
+    ]
+    ++ lib.map (pcr: "--pcr=${toString pcr}") cfg.measuredBoot.pcrs
+  );
 in
 {
   imports = [
@@ -248,6 +291,114 @@ in
         '';
       };
     };
+
+    measuredBoot = {
+      enable = lib.mkEnableOption "Measured Boot";
+
+      pcrs = lib.mkOption {
+        type = lib.types.listOf (
+          lib.types.enum [
+            0
+            1
+            2
+            3
+            4
+            7
+          ]
+        );
+        default = [ ];
+        description = ''
+          PCRs to lock via systemd-pcrlock.
+        '';
+      };
+
+      pcrlockDirectory = lib.mkOption {
+        type = lib.types.path;
+        default = "/var/lib/pcrlock.d";
+        description = ''
+          Directory to store the pcrlock files in.
+        '';
+      };
+
+      pcrlockPolicy = lib.mkOption {
+        type = lib.types.path;
+        default = "/var/lib/systemd/pcrlock.json";
+        description = ''
+          Location to store the pcrlock policy in.
+        '';
+      };
+
+      upstreamStaticMeasurements = lib.mkOption {
+        type = lib.types.listOf lib.types.str;
+        default = [ ];
+        description = ''
+          Filenames of static pcrlock measurements to include from the systemd
+          package.
+        '';
+      };
+
+      staticMeasurements = lib.mkOption {
+        default = { };
+        description = ''
+          Static systemd-pcrlock measurements.
+        '';
+        type = lib.types.attrsOf (
+          lib.types.submodule (
+            {
+              name,
+              config,
+              options,
+              ...
+            }:
+            {
+              options = {
+                source = lib.mkOption {
+                  type = lib.types.path;
+                  description = "Path of the source file.";
+                };
+                json = lib.mkOption {
+                  default = null;
+                  type = lib.types.nullOr json.type;
+                  description = ''
+                    systemd-pcrlock components in their literal form. This option is directly transformed to a JSON.
+                  '';
+                };
+              };
+              config = {
+                source = lib.mkIf (config.json != null) (
+                  lib.mkDerivedConfig options.json (json.generate "${name}.pcrlock")
+                );
+              };
+            }
+          )
+        );
+      };
+
+      autoCryptenroll = {
+        enable = lib.mkEnableOption "automatically re-enroll systemd-pcrlock TPM2 policy into LUKS volume";
+
+        device = lib.mkOption {
+          type = lib.types.nullOr lib.types.str;
+          default = null;
+          description = ''
+            The device that is encrypted via LUKS2 to enroll the TPM2 policy into.
+
+            This is useful for unattended systems to upgrade a LUKS2 volume
+            from being locked against a static PCR to a full systemd-pcrlock
+            policy.
+          '';
+        };
+
+        autoReboot = lib.mkEnableOption "" // {
+          description = ''
+            Whether to automatically reboot after preparing the measurements.
+
+            Enable this to enroll the new systemd-pcrlock policy with full
+            protection without having to wait for a manual reboot.
+          '';
+        };
+      };
+    };
   };
 
   config = lib.mkIf cfg.enable {
@@ -262,6 +413,16 @@ in
             2. Accept the risk via autoEnrollKeys.allowBrickingMyMachine
         '';
       }
+      {
+        assertion = cfg.measuredBoot.enable -> (configurationLimit > 0 && configurationLimit <= 8);
+        message = ''
+          If Measured Boot is enabled, you cannot store more than 8 generations on the ESP.
+
+            This is a strict limit required and enforced by systemd-pcrlock.
+
+            Set `boot.lanzaboote.configurationLimit = 8;` to reduce the number of generations you store.
+        '';
+      }
     ];
 
     boot.bootspec = {
@@ -274,15 +435,67 @@ in
     boot.loader.external = {
       enable = true;
       installHook = pkgs.writeShellScript "bootinstall" (
-        lib.concatStringsSep "\n" (map mkInstallCommand efiSysMountPoints)
+        ''
+          ${lib.concatStringsSep "\n" (map mkInstallCommand efiSysMountPoints)}
+        ''
+        + lib.optionalString cfg.measuredBoot.enable ''
+          echo "Predicting the PCR state for future boots..."
+          ${makePolicyCommand}
+        ''
       );
     };
+    boot.lanzaboote.measuredBoot.upstreamStaticMeasurements =
+      lib.optionals (pcr 0 || pcr 1 || pcr 2 || pcr 3 || pcr 4) [
+        "500-separator.pcrlock.d/300-0x00000000.pcrlock"
+      ]
+      ++ lib.optionals (pcr 4) [
+        "350-action-efi-application.pcrlock"
+      ]
+      ++ lib.optionals (pcr 7) [
+        "400-secureboot-separator.pcrlock.d/300-0x00000000.pcrlock"
+      ];
 
     environment.etc."sbctl/sbctl.conf" =
       lib.mkIf (cfg.autoGenerateKeys.enable || cfg.autoEnrollKeys.enable)
         {
           source = sbctlConfigFile;
         };
+
+    # Write this to /etc so that manually calling systemd-pcrlock by the user
+    # still works without them having to specify the directory.
+    environment.etc."pcrlock" = lib.mkIf cfg.measuredBoot.enable {
+      target = "pcrlock.d";
+      source = staticMeasurements;
+    };
+
+    systemd.additionalUpstreamSystemUnits = lib.mkIf cfg.measuredBoot.enable [
+      "systemd-pcrlock-make-policy.service"
+      "systemd-pcrlock-firmware-code.service"
+      "systemd-pcrlock-firmware-config.service"
+      "systemd-pcrlock-secureboot-policy.service"
+      "systemd-pcrlock-secureboot-authority.service"
+    ];
+    systemd.services.systemd-pcrlock-make-policy = lib.mkIf cfg.measuredBoot.enable {
+      wantedBy = [ "sysinit.target" ];
+
+      serviceConfig.ExecStart = [
+        "" # unset previous value
+        makePolicyCommand
+      ];
+    };
+    systemd.targets.sysinit = lib.mkIf cfg.measuredBoot.enable {
+      wants =
+        lib.optionals (pcr 0 || pcr 2) [
+          "systemd-pcrlock-firmware-code.service"
+        ]
+        ++ lib.optionals (pcr 1 || pcr 3) [
+          "systemd-pcrlock-firmware-config.service"
+        ]
+        ++ lib.optionals (pcr 7) [
+          "systemd-pcrlock-secureboot-policy.service"
+          "systemd-pcrlock-secureboot-authority.service"
+        ];
+    };
 
     systemd.services.generate-sb-keys = lib.mkIf cfg.autoGenerateKeys.enable {
       wantedBy = [ "multi-user.target" ];
@@ -316,13 +529,11 @@ in
           "!${espMountPoint}/loader/keys/auto/KEK.auth"
           "!${espMountPoint}/loader/keys/auto/db.auth"
         ];
-        SuccessAction = lib.mkIf cfg.autoEnrollKeys.autoReboot "reboot";
       };
 
       serviceConfig = {
         Type = "oneshot";
-        # SuccessAction doesn't trigger if the service is RemainAfterExit
-        RemainAfterExit = lib.mkIf (!cfg.autoEnrollKeys.autoReboot) true;
+        RemainAfterExit = true;
         RuntimeDirectory = "prepare-sb-auto-enroll";
         WorkingDirectory = "/run/prepare-sb-auto-enroll";
       };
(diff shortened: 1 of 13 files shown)


blitz:
Looks really good! Will approve once I got a chance to test this on my system. :+1:

frederictobiasc:
Cool! Is there already a way to test this without secureboot?

IdanDor:
Wonderful! I was just hoping to try to figure out how to achieve this on my machine (with probably way too many vulnerabilities and problems that I would open myself up to).

Could you clarify the used PCR's meaning, and why are only they needed?
I'm not an expert, especially on lanzaboote image, but it seems to me that kernel command line (PCR 8?) and kernel image (PCR 9?) are important.
Are they part of PCR 4 under this setup? Or perhaps they should be none default as they would require a lot?
Similarly, PCRs 1 and 2 seem like relevant optional PCRs for paranoid people?

Another question, is what is different between this and the open PR of https://github.com/nix-community/lanzaboote/pull/551, which seems to do this rather differently?

Finally, in my short research I have stumbled upon [this common misconfiguration](https://oddlama.org/blog/bypassing-disk-encryption-with-tpm2-unlock/) which seems relevant here, and would allow extraction of automatic LUKS TPM2 key unlock (password-less) in case of physical access.
Did I correctly understand this? Can this be easily fixed here or should a warning be added to the documentation?

arianvp (commented):
The eval error doesn't tell me how to fix the 8 generation limit

arianvp:
```
error: builder for '/nix/store/5z714xg8zhjwsmfw2ybqnsp9daj1lsld-system-units.drv' failed with exit code 1;
       last 1 log lines:
       > ln: failed to create symbolic link '/nix/store/jdw8jd9lqh0hw8nq76lc53sfcx0wsxwd-system-units/systemd-pcrlock-make-policy.service': File exists
       For full logs, run:
         nix log /nix/store/5z714xg8zhjwsmfw2ybqnsp9daj1lsld-system-units.drv
error: 1 dependencies of derivation '/nix/store/cynx2mw1y6rwxsilav09ks529k9i7b4y-etc.drv' failed to build
error: 1 dependencies of derivation '/nix/store/l94abd5chxajgc5agh1bnd4mr9fqiyki-activate.drv' failed to build
error: 1 dependencies of derivation '/nix/store/pslh52ay869n34gs6n7xidndcx2k1xpj-nixos-system-framework-26.05.20260401.6201e20.drv' failed to build
Command 'nix --extra-experimental-features 'nix-command flakes' build --print-out-paths '.#nixosConfigurations."framework".config.system.build.toplevel'' returned non-zero exit status 1.
```
hmm

nix/modules/lanzaboote.nix
```diff
@@ -263,16 +425,72 @@ in
...
+      '';
+    };
+
+    systemd.additionalUpstreamSystemUnits = lib.mkIf cfg.measuredBoot.enable [
+      "systemd-pcrlock-make-policy.service"
+      "systemd-pcrlock-firmware-code.service"
+      "systemd-pcrlock-secureboot-policy.service"
+      "systemd-pcrlock-secureboot-authority.service"
+    ];
+    systemd.services."systemd-pcrlock-make-policy" = lib.mkIf cfg.measuredBoot.enable {
```
arianvp: This should be `sysinit.wants` otherwise you get empty bogus units

nikstur:
> ```
> error: builder for '/nix/store/5z714xg8zhjwsmfw2ybqnsp9daj1lsld-system-units.drv' failed with exit code 1;
>        last 1 log lines:
>        > ln: failed to create symbolic link '/nix/store/jdw8jd9lqh0hw8nq76lc53sfcx0wsxwd-system-units/systemd-pcrlock-make-policy.service': File exists
>        For full logs, run:
>          nix log /nix/store/5z714xg8zhjwsmfw2ybqnsp9daj1lsld-system-units.drv
> error: 1 dependencies of derivation '/nix/store/cynx2mw1y6rwxsilav09ks529k9i7b4y-etc.drv' failed to build
> error: 1 dependencies of derivation '/nix/store/l94abd5chxajgc5agh1bnd4mr9fqiyki-activate.drv' failed to build
> error: 1 dependencies of derivation '/nix/store/pslh52ay869n34gs6n7xidndcx2k1xpj-nixos-system-framework-26.05.20260401.6201e20.drv' failed to build
> Command 'nix --extra-experimental-features 'nix-command flakes' build --print-out-paths '.#nixosConfigurations."framework".config.system.build.toplevel'' returned non-zero exit status 1.
> ```
> 
> hmm

Adjusted the message to include a concrete solution.

arianvp:
I get in the logs:

```
Unable to recognize 5 components in event log.
Event log record 15 (PCR 1, "Raw: \001\000\000\000\000\000\000\000D\025\375\362\224\227,J\231.\345\273\317 \343\224\000\020\236C\000\000\000\000") not matching any component.
Event log record 29 (PCR 4, "File: \EFI\systemd\systemd-bootx64.efi") not matching any component.
Event log record 28 (PCR 5, "GPT: disk 8e57f999-27fb-df46-bab4-023a48b666c8") not matching any component.
Event log record 41 (PCR 9, "Linux: kernel command line") not matching any component.
Event log record 32 (PCR 11, "String: .linux") not matching any component.
Event log record 40 (PCR 12, "String: Global credentials initrd") not matching any component.
Event log record 46 (PCR 15, "cryptsetup:root:31e89180-a13a-4afb-8b36-75d9dac50997") not matching any component.
PCR 0 (platform-code) matches event log and fully consists of recognized measurements. Including in set of PCRs.
PCR 4 (boot-loader-code) event log contains unrecognized measurements. Removing from set of PCRs.
PCR 7 (secure-boot-policy) matches event log and fully consists of recognized measurements. Including in set of PCRs.
PCRs dropped from protection mask: 4 (boot-loader-code)
PCRs in protection mask: 0 (platform-code), 7 (secure-boot-policy)
```

which doesn't sound right?  as it kicks out PCR 4 ? We should generate a pcrlock file for the bootloader; no?

nix/modules/lanzaboote.nix
```diff
@@ -263,16 +423,75 @@ in
...
+      wantedBy = [ "sysinit.target" ];
+
+      serviceConfig.ExecStart = [
+        "" # unset previous value
+        makePolicyCommand
+      ];
+    };
+    systemd.targets.sysinit.wants =
```
WilliButz: Currently the PR introduces an empty `sysinit.target` file to the toplevel closure even if `measuredBoot` is not used yet, because the module defines `wants = []` for the target unconditionally. The file contains only an empty `[Unit]` section. The following fixes it for me:

```suggestion
    systemd.targets.sysinit = lib.mkIf cfg.measuredBoot.enable { wants = ... };
```

Could you couple the `wants` to the feature being enabled, maybe to a non-empty set of PCRs being selected? (haven't checked the implementation yet, not sure if enable would be sufficient)

nikstur: Done

nikstur:
> Cool! Is there already a way to test this without secureboot?

You could simply not enroll the key into your UEFI and enable `allowUnsigned` then you can test it without Secure Boot.

nikstur:
@arianvp 

> I get in the logs:

From where is this? From running nixos-rebuild or from the make-policy service? In any case, you first need to reboot before you can include PCR4 in the policy because the currently booted system will not produce the expected measurements (and pcrlock then just omits the entire PCR).

> We should generate a pcrlock file for the bootloader; no?

We do. PCR4 should be locked on the next boot.

arianvp:
It printed that when running `nixos-rebuild boot`

full log:

```
Not checking switch inhibitors (action = boot)
/var/lib/pcrlock.d/635-lanzaboote.pcrlock.d/257.pcrlock written.
/var/lib/pcrlock.d/635-lanzaboote.pcrlock.d/258.pcrlock written.
/var/lib/pcrlock.d/635-lanzaboote.pcrlock.d/259.pcrlock written.
/var/lib/pcrlock.d/635-lanzaboote.pcrlock.d/260.pcrlock written.
/var/lib/pcrlock.d/635-lanzaboote.pcrlock.d/261.pcrlock written.
/var/lib/pcrlock.d/635-lanzaboote.pcrlock.d/262.pcrlock written.
/var/lib/pcrlock.d/635-lanzaboote.pcrlock.d/263.pcrlock written.
/var/lib/pcrlock.d/635-lanzaboote.pcrlock.d/264.pcrlock written.
Updating systemd-boot...
systemd-boot has not been measured. Creating measurement and re-installing...
/var/lib/pcrlock.d/630-bootloader.pcrlock.d/next.pcrlock written.
Installing /boot/EFI/BOOT/BOOTX64.EFI
Installing /boot/EFI/systemd/systemd-bootx64.efi
Collecting garbage...
Successfully installed Lanzaboote.
Predicting the PCR state for future boots...
Couldn't find component '630-bootloader' in event log.
Couldn't find component '630-systemd-boot' in event log.
Couldn't find component '635-lanzaboote' in event log.
Couldn't find component '680-kernel' in event log.
Couldn't find component '720-kernel-initrd' in event log.
Didn't find component '950-shutdown' in event log, assuming system hasn't reached it yet.
Didn't find component '990-final' in event log, assuming system hasn't reached it yet.
Skipped 2 components (950-shutdown, 990-final).
Unable to recognize 5 components in event log.
Event log record 15 (PCR 1, "Raw: \001\000\000\000\000\000\000\000D\025\375\362\224\227,J\231.\345\273\317 \343\224\000\020\236C\000\000\000\000") not matching any component.
Event log record 29 (PCR 4, "File: \EFI\systemd\systemd-bootx64.efi") not matching any component.
Event log record 28 (PCR 5, "GPT: disk 8e57f999-27fb-df46-bab4-023a48b666c8") not matching any component.
Event log record 41 (PCR 9, "Linux: kernel command line") not matching any component.
Event log record 32 (PCR 11, "String: .linux") not matching any component.
Event log record 40 (PCR 12, "String: Global credentials initrd") not matching any component.
Event log record 46 (PCR 15, "cryptsetup:root:31e89180-a13a-4afb-8b36-75d9dac50997") not matching any component.
PCR 0 (platform-code) matches event log and fully consists of recognized measurements. Including in set of PCRs.
PCR 4 (boot-loader-code) event log contains unrecognized measurements. Removing from set of PCRs.
PCR 7 (secure-boot-policy) matches event log and fully consists of recognized measurements. Including in set of PCRs.
PCRs dropped from protection mask: 4 (boot-loader-code)
PCRs in protection mask: 0 (platform-code), 7 (secure-boot-policy)
Predicted future PCRs in 30.952ms.
Retrieved PIN from TPM2 in 742.389ms.
Calculated new PCR policy in 47us.
Updated NV index in 251.948ms.
Written new policy to '/var/lib/systemd/pcrlock.json' and digest to TPM2 NV index 0x189c6d1.
Written new boot policy to '/boot/loader/credentials/pcrlock.bb9e3c32f25d42fb9996c2554c85a604.cred'.
Overall time spent: 1.375919s
Done. The new configuration is /nix/store/px86nbfwsq0fvcmyhj2fcsgw021ixl7j-nixos-system-framework-26.05.20260401.6201e20
```

RaitoBezarius:
Few remarks after testing:

- some firmwares might not be compliant with the PCR4 EFI action event (mine) and a quirk system might make sense.
- there's a weird lifecycle thing where the make-policy is ran too early compared to the set of static measurements config files provided to pcrlock, if this cannot be fixed, perhaps this should be documented?
- of course, weird encryption mechanisms like ZFS native encryption are not supported and it should be documented (aka you're on your own to bind the TPM2 to that)

nikstur:
> Could you clarify the used PCR's meaning, and why are only they needed?
I'm not an expert, especially on lanzaboote image, but it seems to me that kernel command line (PCR 8?) and kernel image (PCR 9?) are important.
Are they part of PCR 4 under this setup? Or perhaps they should be none default as they would require a lot?
Similarly, PCRs 1 and 2 seem like relevant optional PCRs for paranoid people?

I added an explanation document that hopefully answers some of your questions. Most importantly, PCR4 covers both the boot loader and all the boot artifacts. This works for the same reason Secure Boot works in Lanzaboote: because the stub checks hashes for all the involved artifacts. PCR 8 and 9 are included in 4 for this reason.

In the newest changes, I also include the ability to lock against PCR 1,2,3 so you can lock against these as well.

> Another question, is what is different between this and the open PR of https://github.com/nix-community/lanzaboote/pull/551, which seems to do this rather differently?

Ultimately, I want to support both. They are not mutually exclusive. However, I believe this approach is the most important one because it covers all aspects of the boot and not only the UKI.

> Finally, in my short research I have stumbled upon [this common misconfiguration](https://oddlama.org/blog/bypassing-disk-encryption-with-tpm2-unlock/) which seems relevant here, and would allow extraction of automatic LUKS TPM2 key unlock (password-less) in case of physical access.
Did I correctly understand this? Can this be easily fixed here or should a warning be added to the documentation?

Yes you can circumvent this with a pin which is what I recommend in the guide. It's hard to point out all the security pitfalls. Thus we try to avoid including all of them in our docs. Otherwise they become hard to read.

nikstur:
> It printed that when running nixos-rebuild boot

This is expected then. You will need to reboot first before this will include the desired PCRs because your current event log. However you can also just enroll this policy already and it will update automatically on the next boot.

nikstur:
> some firmwares might not be compliant with the PCR4 EFI action event (mine) and a quirk system might make sense.

For you this would already be solved if you could disable one of the upstream measurements. I'd prefer making this configurable instead of adding a fully fledged quirk system.

> there's a weird lifecycle thing where the make-policy is ran too early compared to the set of static measurements config files provided to pcrlock, if this cannot be fixed, perhaps this should be documented?

This should be solved now with the latest changes as we give pcrlock the store path directly without needing it to be in /etc.

> of course, weird encryption mechanisms like ZFS native encryption are not supported and it should be documented (aka you're on your own to bind the TPM2 to that)

The first sentence of the guide is "This guide will walk you through enabling measured boot on a system that
already has some form of `LUKS2` disk encryption and migrating this `LUKS2`
volume to measured boot."

Does this need to be more explicit?

WilliButz:
Nice work! Successfully enrolled a tpm2-token into my luks volume with this :partying_face: 

```
Apr 06 00:52:26 systemd[1]: Starting Make TPM PCR Policy...
Apr 06 00:52:26 systemd-pcrlock[1731]: Event log record 46 (PCR 5, "GPT: disk c125fb91-335b-4c80-8ef0-3abe060d2d0b") not matching any component.
Apr 06 00:52:26 systemd-pcrlock[1731]: Event log record 59 (PCR 9, "Linux: kernel command line") not matching any component.
Apr 06 00:52:26 systemd-pcrlock[1731]: Event log record 50 (PCR 11, "String: .linux") not matching any component.
Apr 06 00:52:26 systemd-pcrlock[1731]: Event log record 58 (PCR 12, "String: Global credentials initrd") not matching any component.
Apr 06 00:52:26 systemd-pcrlock[1731]: PCR 0 (platform-code) matches event log and fully consists of recognized measurements. Including in set of PCRs.
Apr 06 00:52:26 systemd-pcrlock[1731]: PCR 1 (platform-config) matches event log and fully consists of recognized measurements. Including in set of PCRs.
Apr 06 00:52:26 systemd-pcrlock[1731]: PCR 2 (external-code) matches event log and fully consists of recognized measurements. Including in set of PCRs.
Apr 06 00:52:26 systemd-pcrlock[1731]: PCR 3 (external-config) matches event log and fully consists of recognized measurements. Including in set of PCRs.
Apr 06 00:52:26 systemd-pcrlock[1731]: PCR 4 (boot-loader-code) matches event log and fully consists of recognized measurements. Including in set of PCRs.
Apr 06 00:52:26 systemd-pcrlock[1731]: PCR 7 (secure-boot-policy) matches event log and fully consists of recognized measurements. Including in set of PCRs.
Apr 06 00:52:26 systemd-pcrlock[1731]: PCRs in protection mask: 0 (platform-code), 1 (platform-config), 2 (external-code), 3 (external-config), 4 (boot-loader-code), 7 (secure-boot-policy)
Apr 06 00:52:26 systemd-pcrlock[1731]: Predicted future PCRs in 132us.
Apr 06 00:52:27 systemd-pcrlock[1731]: Calculated new PCR policy in 26us.
Apr 06 00:52:27 systemd-pcrlock[1731]: Updated NV index in 550.210ms.
Apr 06 00:52:27 systemd-pcrlock[1731]: Sealed PIN in 165.810ms.
Apr 06 00:52:27 systemd-pcrlock[1731]: Written new policy to '/var/lib/systemd/pcrlock.json' and digest to TPM2 NV index 0x1b26e72.
Apr 06 00:52:27 systemd-pcrlock[1731]: Written new boot policy to '/efi/loader/credentials/pcrlock.2c1bad5cafec4ea68c80eeff0dfce9cb.cred'.
Apr 06 00:52:27 systemd-pcrlock[1731]: Overall time spent: 954.063ms
Apr 06 00:52:27 systemd[1]: Finished Make TPM PCR Policy.
```

I just added this config and went through the documented steps:
```nix
{
  boot.lanzaboote = {
    measuredBoot = {
      enable = true;
      pcrs = [ 0 1 2 3 4 7 ];
    };
  };
}
```

RaitoBezarius:
> For you this would already be solved if you could disable one of the upstream measurements. I'd prefer making this configurable instead of adding a fully fledged quirk system.

Works for me.

> > there's a weird lifecycle thing where the make-policy is ran too early compared to the set of static measurements config files provided to pcrlock, if this cannot be fixed, perhaps this should be documented?

> This should be solved now with the latest changes as we give pcrlock the store path directly without needing it to be in /etc.

Awesome.

> Does this need to be more explicit?

I think making it clear that ZFS native encryption is not LUKS2 will be helpful (I know it isn't without it being written but something something about being distracted :P).

IdanDor:
Good answers, thank you. I liked the change to "you should use a pin in case of attended upgrades...".

What is the expected update flow here for tpm+pin?
If I understand correctly, `autoCryptenroll` is for tpm-only update, but I didn't find in the systemd man pages I read, a mechanism when, after an update, I enter a recovery password, and the luks-tpm-with-pin automatically corrects itself, yes?
Or perhaps are we supposed to "reinstall measured boot" after any boot update?

RaitoBezarius (approved):

nix/modules/lanzaboote.nix
```diff
@@ -22,19 +22,69 @@ let
...
+  '';
+
+  makePolicyCommand = lib.escapeShellArgs (
+    [
+      "${config.systemd.package}/lib/systemd/systemd-pcrlock"
+      "make-policy"
+      "--components=${staticMeasurements}"
+      "--components=/etc/pcrlock.d/"
+      "--components=/run/pcrlock.d/"
+      "--components=/var/lib/pcrlock.d/"
```
Ma27: Shouldn't this be `cfg.pcrlockDirectory`?

nikstur:
> What is the expected update flow here for tpm+pin?

I added some clarification to the getting started guide about this. You will not need to do anything. Lanzaboote will take care of creating new measurements and updating the policy. You enroll the pin once via `systemd-cryptenroll` and will not have to touch it after.

nix/modules/lanzaboote.nix
```diff
@@ -262,6 +413,16 @@ in
             2. Accept the risk via autoEnrollKeys.allowBrickingMyMachine
         '';
       }
+      {
+        assertion = cfg.measuredBoot.enable -> (configurationLimit > 0 && configurationLimit <= 8);
```
Ma27, line 417: Do you know if there's a reason for this limit or is it arbitrary? I wonder if we should convince upstream into increasing this number, I found higher amounts of generations valuable on some of my machines in the past.

Ma27 (commented):
Successfully tested this on a Framework 13 laptop (first generation) running NixOS 25.11 with systemd 258.3.

---

Without having dug deep into the issue, PCR1 is flaky on my machine which means I got an error like this from systemd-pcrlock after rebooting with PCR1 changing:

```
[snens] WARNING:esys:src/tss2-esys/api/Esys_PolicyAuthorizeNV.c:311:Esys_PolicyAuthorizeNV_Finish() Received TPM Error 
[snens] ERROR:esys:src/tss2-esys/api/Esys_PolicyAuthorizeNV.c:104:Esys_PolicyAuthorizeNV() Esys Finish ErrorCode (0x00000084) 
[snens] Failed to submit AuthorizeNV policy: Remote address changed
[snens] Failed to install bootloader
```

Aaccording to systemd sources this happens when predictions don't match actual measurements. However, systemd-pcrlock gives me a check-mark for the H/R/C column in the measurement table. The way I found out was by doing `mv pcrlock.json pcrlock.json.bak`, letting lanzaboote/systemd-pcrlock generate a new pcrlock.json and then diffing which predictions have changed.

I'm not really into measured-boot, but having systemd-pcrlock showing me that everything is fine is not what I'd expect, do you think this is a systemd bug? If not, is there a better way to debug this problem than what I did? Happy to contribute a how-to guide :) 

Additionally, this caused a second problem: the error persisted after removing `1` from `boot.lanzaboote.measuredBoot.pcrs`. The error above only went away after deleting everything in `/var/lib/pcrlock.d` & `/var/lib/systemd/pcrlock.json`, installing a new bootloader and rebooting once.

---

Finally, one question for my own understanding (and I'm not very deep into this topic at all as you may know): previously I've been doing the "we have measured-boot at home"-style setup of locking systemd-cryptenroll against pcr15 being a nulled string and then measuring the volume key into pcr15 to make sure the tpm's state has changed before entering stage2 such that the luks key ins't available to stage2.

Is this something systemd-pcrlock is taking care of by its own here in a way I'm not seeing? Because the PCRs we're measuring don't change after leaving initrd, no?

nix/modules/lanzaboote.nix
Ma27: Oh you fixed this in the meantime, cool ;-)

nikstur:
> I'm not really into measured-boot, but having systemd-pcrlock showing me that everything is fine is not what I'd expect, do you think this is a systemd bug? If not, is there a better way to debug this problem than what I did? Happy to contribute a how-to guide :)
> 
> Additionally, this caused a second problem: the error persisted after removing 1 from boot.lanzaboote.measuredBoot.pcrs. The error above only went away after deleting everything in /var/lib/pcrlock.d & /var/lib/systemd/pcrlock.json, installing a new bootloader and rebooting once.

Not sure if this is related to PCR1 being unstable. I've observed this same issue when I re-enrolled my Secure Boot keys. Really not sure what the root cause is because the systemd-pcrlock errors are very cryptic. It would definitely be nice if we added a little guide to our troubleshooting page and open an upstream issue. 

> Is this something systemd-pcrlock is taking care of by its own here in a way I'm not seeing? Because the PCRs we're measuring don't change after leaving initrd, no?

No pcrlock is not taking care of this. How would you update this policy afterwards? In my understanding, this would force you to update the pcrlock policy before unlocking your LUKS volume (i.e. in your initrd).

Eventually I want to support both pcrlock and signed policies. With signed policies we can use the pcrphases to only unlock the secret in the initrd (because the signed policy is embedded in the UKI and thus has a different lifecycle than the pcrlock policy). With only pcrlock we cannot really use this mechanism.

nix/modules/lanzaboote.nix
nikstur: https://github.com/systemd/systemd/issues/41526

Already reported upstream. It currently is a hard requirement from systemd-pcrlock but it can be changed upstream.

blitz (approved):

nikstur:
Last few changes were just adapting the docs to be a little more clear. Importantly, added a changelog entry.

Merging this now as it's running on a number of people's machines now.

Stebalien:
>> Is this something systemd-pcrlock is taking care of by its own here in a way I'm not seeing? Because the PCRs we're measuring don't change after leaving initrd, no?

> No pcrlock is not taking care of this. How would you update this policy afterwards? In my understanding, this would force you to update the pcrlock policy before unlocking your LUKS volume (i.e. in your initrd).

The value of PRC 15 before unlocking the root volume is always 0, as far as I know. Can it not be hard-coded to 0 in pcrlock.json? And/or is it not possible to pass an additional `--tpm2-pcrs=15:sha256=0000...` to `systemd-cryptenroll`?

WilliButz:
> And/or is it not possible to pass an additional `--tpm2-pcrs=15:sha256=0000...` to `systemd-cryptenroll`?

Yes, it is possibe to combine the pcrlock policy with static PCR values like that

Stebalien:
Ah, the downside to that approach is that auto-cryptenroll won't work because it uses `--unlock-tpm2-device`. The entire point of locking the boot device to PCR 15=0x000... is to prevent the boot device's encryption key from being extracted from the TPM after unlocking the first device.

PR #566, opened 2026-04-06
fix: don't reinstall systemd-boot when pcrlock is unconfigured

Titaniumtown:
I discovered a flakey test, `keep_systemd_boot_binaries`. This is because of a bug introduced by #564. The `measurement_exists` check was faulty, causing constant overwrites. Inverting the semantics of `measurement_exists` -> `measurement_missing`, fixes this issue due to the nature of `Option::is_some_and`.

The test was flakey because of the assertion being done used mtime(), which has second-level resolution. So the overwrite was only caught sometimes based on timing. For me, this caused really weird behavior, where Lanzaboote would would compile and all checks would pass on one machine, but not another due to the difference in performance, this is especially the case when under load.

Root caused with Claude Opus 4.6 via Oh-My-Pi

### DIFF
--- a/rust/tool/systemd/src/install.rs
+++ b/rust/tool/systemd/src/install.rs
@@ -453,17 +453,17 @@ impl<S: Signer> Installer<S> {
             log::warn!("systemd-boot is not signed. Replacing it with a signed binary...")
         };
 
-        let measurement_exists = self
+        let measurement_missing = self
             .pcrlock_paths
             .as_ref()
-            .is_some_and(|p| p.bootloader_measurement("current").exists());
-        if !measurement_exists {
+            .is_some_and(|p| !p.bootloader_measurement("current").exists());
+        if measurement_missing {
             log::warn!(
                 "systemd-boot has not been measured. Creating measurement and re-installing..."
             )
         };
 
-        if newer_systemd_boot_available || !systemd_boot_is_signed || !measurement_exists {
+        if newer_systemd_boot_available || !systemd_boot_is_signed || measurement_missing {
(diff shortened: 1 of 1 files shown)


RaitoBezarius:
Hello, the lanzaboote project does not accept AI authored changes, you can find the policy we adopted in https://docs.afnix.fr/policies/ai.html.

I will close it in the meantime, reading the changes, the explanations: I feel like AI tools were used without reviewing their contents and verifying this fixed an issue.

Feel free to resubmit a PR that you authored yourself or an issue we can help with. Thank you for your understanding.

Titaniumtown:
AI was not used to author the changes. I said that it was used to root cause the issue and I wrote the fix.

Titaniumtown:
If you have questions about the issue or PR, please pose them.

nikstur:
Indeed, the problem you uncovered was correct. Thank your for that!

However, your description was quite verbose and with the odd details of what AI tools you used to uncover it, it seemed like you were trying to propose an AI contribution which is prohibited by our policy. Please excuse this oversight on our side.

Titaniumtown:
I'm sorry about my weird description. I hate writing PR descriptions! It's my least favorite part of the process. Thank you for apologizing, I was very taken aback by the response.

> with the odd details of what AI tools you used to uncover it

I just try to be upfront about it because I know it's important to some people.

All I really used AI for was my homeserver's nix derivation was failing and I asked it to determine why it was failing, and it came back to me saying it was because of that flakey test and that file being overridden because of the value of `measurement_exists` being faulty. Hence me using the phrase "root cause".

In the end, I am just glad to see the issue being resolved. Thanks for y'alls hard work! Much appreciated.

PR #568, opened 2026-04-06
stub: remove features that depend on Secure Boot status

nikstur:
This is done in preparation for running Measured Boot without Secure Boot.

Because we measure only the stub into PCR4, we expect that the stub correctly verifies the embedded hashes for the kernel and initrd. Until now, however, the stub would only do this if Secure Boot was enabled.

### DIFF
--- a/rust/uefi/stub/src/common.rs
+++ b/rust/uefi/stub/src/common.rs
@@ -6,6 +6,7 @@ use uefi::{
 };
 
 use linux_bootloader::linux_loader::InitrdLoader;
+use linux_bootloader::measure::measure_boot_loader;
 use linux_bootloader::pe_loader::Image;
 use linux_bootloader::pe_section::pe_section_as_string;
 
@@ -20,28 +21,46 @@ pub fn extract_string(pe_data: &[u8], section: &str) -> uefi::Result<CString16> 
 ///
 /// If Secure Boot is active, this is always the embedded one (since the one passed from the bootloader may come from a malicious type 1 entry).
 /// If Secure Boot is not active, the command line passed from the bootloader is used, falling back to the embedded one.
-pub fn get_cmdline(embedded: &CStr16, secure_boot_enabled: bool) -> Vec<u8> {
-    if secure_boot_enabled {
-        // The command line passed from the bootloader cannot be trusted, so it is not used when Secure Boot is active.
-        embedded.as_bytes().to_vec()
-    } else {
-        let passed = boot::open_protocol_exclusive::<LoadedImage>(boot::image_handle())
-            .map(|loaded_image| loaded_image.load_options_as_bytes().map(|b| b.to_vec()));
-        match passed {
-            Ok(Some(passed)) => passed,
-            // If anything went wrong, fall back to the embedded command line.
-            _ => embedded.as_bytes().to_vec(),
+///
+/// If we read a user provided cmdline, we measure it into PCR4 to invalidate the PCR. This way we
+/// can support Measured Boot without Secure Boot while still giving users the option to edit the
+/// kernel cmdline on the fly for debugging.
+pub fn get_cmdline(embedded: &CStr16) -> Vec<u8> {
+    let secure_boot_enabled = get_secure_boot_status();
+    if !secure_boot_enabled {
+        let load_options = get_load_options();
+        if let Some(passed) = load_options {
+            // The passed cmdline is allowed to be the same as the embedded one without needing to
+            // be measured.
+            if passed == embedded.as_bytes() {
+                return passed;
+            }
+            // Measuring is mandatory here. If the measurement fails, return the embedded cmdline.
+            if measure_boot_loader(&passed, "Custom user provided kernel cmdline").is_ok() {
+                return passed;
+            }
         }
     }
+    embedded.as_bytes().to_vec()
 }
 
-/// Check whether Secure Boot is active, and we should be enforcing integrity checks.
+/// Obtain the load options of the currently loaded image.
+///
+/// This is used for example to pass a customer user provided kernel cmdline from the boot loader
+/// to the booted image.
+fn get_load_options() -> Option<Vec<u8>> {
+    boot::open_protocol_exclusive::<LoadedImage>(boot::image_handle())
+        .ok()
+        .and_then(|loaded_image| loaded_image.load_options_as_bytes().map(|b| b.to_vec()))
+}
+
+// Check whether Secure Boot is active.
 ///
 /// In case of doubt, true is returned to be on the safe side.
 pub fn get_secure_boot_status() -> bool {
     // The firmware initialized SecureBoot to 1 if performing signature checks, and 0 if it doesn't.
     // Applications are not supposed to modify this variable (in particular, don't change the value from 1 to 0).
-    let secure_boot_enabled = runtime::get_variable(
+    runtime::get_variable(
         cstr16!("SecureBoot"),
         &VariableVendor(guid!("8be4df61-93ca-11d2-aa0d-00e098032b8c")),
         &mut [1],
@@ -64,13 +83,7 @@ pub fn get_secure_boot_status() -> bool {
             warn!("Failed to read SecureBoot variable: {err}. Performing verification anyway.");
             true
         }
-    });
-
-    if !secure_boot_enabled {
-        warn!("Secure Boot is not active!");
-    }
-
-    secure_boot_enabled
+    })
 }
 
 /// Boot the Linux kernel without checking the PE signature.
--- a/rust/uefi/stub/src/thin.rs
+++ b/rust/uefi/stub/src/thin.rs
@@ -58,8 +58,12 @@ impl UkiComponents {
             .expect("Failed to read initrd file into memory");
 
         let secure_boot_enabled = get_secure_boot_status();
-        check_hash(&kernel_data, kernel_hash, "Kernel", secure_boot_enabled)?;
-        check_hash(&initrd_data, initrd_hash, "Initrd", secure_boot_enabled)?;
+        if !secure_boot_enabled {
+            warn!("Secure Boot is not active!");
+        }
+
+        check_hash(&kernel_data, kernel_hash, "Kernel")?;
+        check_hash(&initrd_data, initrd_hash, "Initrd")?;
 
         Ok(Self {
             kernel_data,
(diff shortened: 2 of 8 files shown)


rust/uefi/stub/src/thin.rs
```diff
@@ -58,8 +58,12 @@ impl UkiComponents {
...
 
         let secure_boot_enabled = get_secure_boot_status();
-        check_hash(&kernel_data, kernel_hash, "Kernel", secure_boot_enabled)?;
-        check_hash(&initrd_data, initrd_hash, "Initrd", secure_boot_enabled)?;
+        if !secure_boot_enabled {
+            warn!("Secure Boot is not active!");
+        }
+
+        check_hash(&kernel_data, kernel_hash, "Kernel")?;
+        check_hash(&initrd_data, initrd_hash, "Initrd")?;
```
blitz, line 66: Technically, we wouldn't need to insist on this in the measured boot without secure boot scenario. We measure the content of kernel and initrd and not their hashes, if I remember correctly. But I haven't seen this being a problem in years, so I'm okay with making this mandatory and simplifying the code flow.

blitz: This is the case since #546.

nikstur: But this is measured into PCR 11, not PCR 4. In PCR 4 we only measure the bytes of the Lanzaboote image and thus only the hashes of the kernel and initrd.

ElvishJerricco: I agree with nikstur; this isn't so much about what is measured *at least somewhere*, because yes the kernel and initrd are both measured elsewhere as well. It's about ensuring they are also measured in PCR 4 as part of the stub binary, just like how they are with systemd-stub. If the hash that is measured as part of the stub binary isn't verified against the file we load, then this isn't happening.

RaitoBezarius:
Counter-proposal:
```diff
diff --git a/rust/uefi/linux-bootloader/src/measure.rs b/rust/uefi/linux-bootloader/src/measure.rs
index c9174ed..15b3475 100644
--- a/rust/uefi/linux-bootloader/src/measure.rs
+++ b/rust/uefi/linux-bootloader/src/measure.rs
@@ -25,6 +25,8 @@ use uefi::{
 /// 2. The (binary) section contents"
 ///
 /// Measurements are made in canonical order, interleaved: section name, section data, next section name, etc.
+pub const TPM_PCR_INDEX_BOOT_LOADER: PcrIndex = PcrIndex(4);
+
 const TPM_PCR_INDEX_KERNEL_IMAGE: PcrIndex = PcrIndex(11);
 /// This is where lanzastub extends the kernel command line and any passed credentials into
 const TPM_PCR_INDEX_KERNEL_CONFIG: PcrIndex = PcrIndex(12);
diff --git a/rust/uefi/stub/src/common.rs b/rust/uefi/stub/src/common.rs
index d6ac766..a881587 100644
--- a/rust/uefi/stub/src/common.rs
+++ b/rust/uefi/stub/src/common.rs
@@ -13,6 +13,11 @@ pub fn extract_string(pe_data: &[u8], section: &str) -> uefi::Result<CString16>
     Ok(CString16::try_from(string.as_str()).map_err(|_| Status::INVALID_PARAMETER)?)
 }
 
+pub fn get_load_options_cmdline() -> Option<Vec<u8>> {
+    boot::open_protocol_exclusive::<LoadedImage>(boot::image_handle())
+        .map(|loaded_image| loaded_image.load_options_as_bytes().map(|b| b.to_vec()));
+}
+
 // Check whether Secure Boot is active.
 ///
 /// In case of doubt, true is returned to be on the safe side.
diff --git a/rust/uefi/stub/src/thin.rs b/rust/uefi/stub/src/thin.rs
index 63c15d0..9aafbad 100644
--- a/rust/uefi/stub/src/thin.rs
+++ b/rust/uefi/stub/src/thin.rs
@@ -1,6 +1,8 @@
-use crate::common::{boot_linux_unchecked, extract_string, get_secure_boot_status};
+use crate::common::{boot_linux_unchecked, extract_string, get_load_options_cmdline, get_secure_boot_status};
 use alloc::vec::Vec;
+use linux_bootloader::measure::TPM_PCR_INDEX_BOOT_LOADER;
 use linux_bootloader::pe_section::pe_section;
+use linux_bootloader::tpm::tpm_log_event_ascii;
 use linux_bootloader::uefi_helpers::PeInMemory;
 use log::{error, warn};
 use sha2::{Digest, Sha256};
@@ -90,7 +92,29 @@ pub fn boot_linux(
     components: UkiComponents,
     dynamic_initrds: Vec<Vec<u8>>,
 ) -> uefi::Result<()> {
-    let cmdline = components.cmdline.as_bytes().to_vec();
+    let secureboot_enabled = get_secure_boot_status();
+    let embedded_cmdline = components.cmdline.as_bytes().to_vec();
+    let cmdline = match get_load_options_cmdline() {
+        Some(load_options) => {
+            if !secureboot_enabled && load_options != embedded_cmdline {
+                // Measure this into PCR4 to prevent a malicious user to obtain an advantage
+                // via PCR policies.
+
+                if tpm_log_event_ascii(
+                    TPM_PCR_INDEX_BOOT_LOADER,
+                    &load_options,
+                    "Divergent LoadOptions"
+                )? {
+                    load_options
+                } else {
+                    embedded_cmdline
+                }
+            } else {
+                embedded_cmdline
+            }
+        }
+        _ => embedded_cmdline
+    };
 
     let mut initrd_data = components.initrd_data;
 ```

**Comment by ElvishJerricco** [comment]
so, my primary motivation in my other comment was that we probably want lanzaboote to match systemd-stub about all this. (I mean, really systemd-stub's behavior should be formalized as a spec like the TCG UEFI measurement spec (no, the Linux TPM PCR Registry is not good enough), so that we could just adhere to it and wiggle within the discretion it allows, but absent that we should probably just do what systemd-stub does exactly when possible).

I think it's *really* valuable to be able to change the cmdline when secure boot is disabled; it's a troubleshooting method. When something isn't working, naturally you can't just use cmdline to gain free root access on the console when secure boot is enabled, but it's important to be able to just disable secure boot (assuming your platform authorizes you to, changing PCR 7), edit the cmdline, and gain a root shell that way, among other things.

As for how to measure things in this case, indulge me in considering the nature of what systemd-stub is doing. There is a difference between the embedded features of the binary, and the external features. The former goes into PCR 11, and the latter goes into PCR 12. But the external features are even subdivided further between addons and user-owned features like credentials and *the overridden cmdline*, despite them all going into PCR 12. What differentiates them in measurement is the fact that *addons* go into PCR 4 via `LoadImage` and friends. The meaningful separation between addons and other PCR 12 components is their signatures; addons are signed, and only considered "user-owned" in the MOK sense, but I think it's useful to differentiate between MOK-owned and user-owned.

So *why* are vendor / MOK-owned components measured in PCR 4, but user-owned ones like the overridden cmdline are not? I think there's a bit of dissonance in how systemd's documentation about this describes it. Obviously there's the mechanical aspect of just being loaded by `LoadImage`, but I think we can fit an actual reason to it as well. User-owned things do not inherently require any kind of authentication to change (though of course their individual use cases might), whereas addons require your vendor or MOK signature. On the one hand, the overridden cmdline is user-owned, like credentials. On the other hand, the user is not *allowed* to override the cmdline when secure boot is enabled because it is too critical to platform security, like initrd. So in a measured-boot-only scenario, which is it? I think that's where it starts to break down.

The cmdline can't be user-owned and simultaneously so critical to platform security that it can't be overridden. So I think in principle the cmdline should be considered vendor / MOK-owned. But in a measured-boot-only scenario, *that distinction doesn't exist*. So in a measured-boot-only scenario, I think *everything is user-owned*, which allows us a bit of a cheat code. You can think of the overridden cmdline not as a user-owned component, but as an ad-hoc, unsigned cmdline addon generated at runtime. It's not allowed with secure boot because it's unsigned, but it *would be* measured like an addon. This gives the *behavior* we want, in terms of being allowed to override the cmdline at runtime, without losing all the conceptual distinctions that matter for secure boot or measured boot.

Of course this is not what systemd-stub is actually doing. The overridden cmdline is only going to PCR 12, while addons go to PCR 4 as well. But my point is that in principle it *should* be going to both, just like addons. In fact, *in principle*, I'm arguing there should be a distinction between addons and user-owned features like credentials, so they should probably not be measured in the same PCR 12, but rather in different ones. And an overridden cmdline goes in the addon ones.

TL;DR: Measure it in both PCR 4 and 12, and yell at systemd-stub that they should too.

**Comment by nikstur** [comment]
Added a measurement of the passed kernel cmdline into PCR4 now. I'm confident now that this allows us to enable Measured Boot without Secure Boot.

**Comment by nikstur** [comment]
The first commit 9bc1e93c2a8e4b12339500cba06f8c46e8209017 cleans up a function that returned a Result<bool>. See the commit message for more detail.

File: rust/uefi/stub/src/common.rs [hunk 2]
```diff
@@ -20,28 +21,42 @@ pub fn extract_string(pe_data: &[u8], section: &str) -> uefi::Result<CString16>
...
+///
+/// If we read a user provided cmdline, we measure it into PCR4 to invalidate the PCR. This way we
+/// can support Measured Boot without Secure Boot while still giving users the option to edit the
+/// kernel cmdline on the fly for debugging.
+pub fn get_cmdline(embedded: &CStr16) -> Vec<u8> {
+    let secure_boot_enabled = get_secure_boot_status();
+    if !secure_boot_enabled {
+        let load_options = get_load_options();
+        // Measuring is mandatory here. If the measurement fails, return the embedded cmdline.
+        if let Some(passed) = load_options
```
RaitoBezarius: If `load_options == embedded`, you should not measure.

nikstur: Anything you want <3

RaitoBezarius (approved):

ElvishJerricco:
What's the counterargument to my suggestion to measure the overridden cmdline into PCR 12 as well?

nikstur:
> What's the counterargument to my suggestion to measure the overridden cmdline into PCR 12 as well?

No counterargument. Not just particularly relevant for the problem I'm trying to solve. PCR12 is not fully implemented in Lanzaboote right now, so I wouldn't want to rely on it. Adding new measurements to PCR12 would be a new feature in my opinion that I wouldn't want to implement in this PR. Additionally (and probably more importantly) [PCR 12 currently cannot be used with pcrlock](https://github.com/systemd/systemd/issues/33546).

PR #571, opened 2026-04-08
Keep the link together

SuperSandro2000:


nikstur:
The line breaks are introduced by a vim plugin which automatically breaks lines. Closing as this is no material improvement.

SuperSandro2000:
That vim plugin does really interesting stuff to enforce short lines.

PR #576, opened 2026-04-10
stub: abstract over a ParsedPe

nikstur:
I believe this will make it a little easier to maintain in the future and get's rid of some duplication.


rust/uefi/linux-bootloader/src/uefi_helpers.rs
```diff
@@ -59,3 +61,67 @@ pub fn booted_image_file() -> Result<PeInMemory> {
...
+    pub fn section_data(&self, section_name: &str) -> Option<Vec<u8>> {
+        self.parsed
+            .sections
+            .iter()
+            .find(|s| s.name().map(|n| n == section_name).unwrap_or(false))
+            .and_then(|s| read_data_from_section_table(self.data, s))
+    }
+
+    /// Extracts the data of a section of a loaded PE image and returns it as a string.
+    pub fn section_data_as_string(&self, section_name: &str) -> Option<String> {
```
RaitoBezarius: I don't understand why we should make this a function given we have `section_data` now, it's literally `section_data().and_then(|d| String::from_utf8(d).ok())` which happens literally in one call site.

RaitoBezarius (approved):

rust/uefi/linux-bootloader/src/uefi_helpers.rs
nikstur: Done

Issue #584, opened 2026-04-16
Measured Boot PCR 7 Prediction Issues (Systemd Issue?)


eth-p:
First of all, thank you for implementing Measured Boot!

I'm running an AMD system and trying to get Measured Boot set up using PCRs 0, 4, and 7 as the [Enable Measured Boot](https://nix-community.github.io/lanzaboote/how-to-guides/enable-measured-boot.html) documentation suggests. Everything works fine when using only PCR 0 and PCR 4, but adding PCR 7 prevents the disk from automatically unlocking.

The relevant `journalctl -b` logs are:

```
systemd-cryptsetup[563]: WARNING:esys:src/tss2-esys/api/Esys_PolicyAuthorizeNV.c:311:Esys_PolicyAuthorizeNV_Finish() Received TPM Error
systemd-cryptsetup[563]: ERROR:esys:src/tss2-esys/api/Esys_PolicyAuthorizeNV.c:104:Esys_PolicyAuthorizeNV() Esys Finish ErrorCode (0x00000084)
systemd-cryptsetup[563]: TPM policy does not match current system state. Either system has been tempered with or policy out-of-date: Remote address changed
```

This looks to be related to https://github.com/systemd/systemd/issues/40381, which is a bug in `systemd-pcrlock predict` predicting the wrong value for PCR 7. Similar to what the author of that issue noticed, my predicted PCR 7 hash stored in `/var/lib/systemd/pcrlock.json` is different from the actual PCR 7 hash returned by `systemd-analyze pcrs`.

```
predicted:         635189b478c269eb122c359aa4e742ade1d77f931ea1155d859e8ead10f1b523
pcrlock.json:      635189b478c269eb122c359aa4e742ade1d77f931ea1155d859e8ead10f1b523
systemd-analyze:   5db797f2809ef15fd4e64e758cfae5a949f120e9b5a1fb9baf824e6921933f4c
```

Without the upstream systemd issue being fixed I'm not sure there's much that you can do about it in Lanzaboote, but do you have any suggestions for how to approach working around this without removing PCR 7 from the measured boot entirely?

---

## System Info

`uname -a`
```
Linux MY-PC 6.19.11-cachyos #1-NixOS SMP PREEMPT_DYNAMIC Tue Jan  1 00:00:00 UTC 1980 x86_64 GNU/Linux
```

Systemd: `systemd 259 (259.3)`

[...]

PR #605, opened 2026-05-24
docs: Add check for systemd PCR 7 prediction bug

eth-p:
Following @nikstur's suggestion from #584, I added information about systemd's PCR 7 prediction bug to the docs.

I opted to add it to the how-to guide rather than the troubleshooting guide, though. The PCR 7 prediction issue isn't be apparent until the first boot after enabling measured boot, and I believe it's better user experience to avoid the issue rather than have to perform troubleshooting after encountering it.

### DIFF
--- a/docs/how-to-guides/enable-measured-boot.md
+++ b/docs/how-to-guides/enable-measured-boot.md
@@ -22,6 +22,27 @@ If this says anything other than `yes`, you will not be able to use Lanzaboote
 for Measured Boot because the TPM in your system is not supported by
 systemd-pcrlock.
 
+### Check PCR 7 Ordering
+
+If your system has an AMD processor and is using the fTPM, a
+[bug in systemd](https://github.com/systemd/systemd/issues/40381) may prevent
+PCR 7 from being predicted correctly.
+
+```console
+$ sudo /run/current-system/systemd/lib/systemd/systemd-pcrlock cel 2>/dev/null | jq -r '.[] | select(.pcr == 7) | .content.event_type'
+EV_EFI_VARIABLE_DRIVER_CONFIG
+EV_EFI_VARIABLE_DRIVER_CONFIG
+EV_EFI_VARIABLE_DRIVER_CONFIG
+EV_EFI_VARIABLE_DRIVER_CONFIG
+EV_EFI_VARIABLE_DRIVER_CONFIG
+EV_EFI_VARIABLE_AUTHORITY
+EV_SEPARATOR
+EV_EFI_VARIABLE_AUTHORITY
+```
+
+If there is a `EV_EFI_VARIABLE_AUTHORITY` before the `EV_SEPARATOR`,
+you will not be able to use PCR 7 for measured boot.
+
 ## Enable Measured Boot in Your Config
 
 > [!NOTE]


eth-p:
On a side note, it is _theoretically_ possible to work around the issue.

The `systemd-pcrlock-secureboot-authority.service` would need to be replaced with something that generates the correct pcrlock.d components from the CEL log:

```bash
mkdir -p /var/lib/pcrlock.d/280-secureboot-authority.pcrlock.d/
mkdir -p /var/lib/pcrlock.d/500-separator.pcrlock.d/
mkdir -p /var/lib/pcrlock.d/620-secureboot-authority.pcrlock.d/

# Before separator.
/run/current-system/systemd/lib/systemd/systemd-pcrlock cel | jq '
    [.[] | select(.pcr == 7)] as $pcr7 
    | ($pcr7.[] | select(.content.event_type == "EV_SEPARATOR")) as $sep 
    | { "records": [
        $pcr7 | .[0:index($sep)]
        | .[] | select(.content.event_type == "EV_EFI_VARIABLE_AUTHORITY")
        | with_entries(select([.key] | inside(["pcr", "digests"])))
    ]}
' >/var/lib/pcrlock.d/280-secureboot-authority.pcrlock.d/generated.pcrlock

# Separator.
/run/current-system/systemd/lib/systemd/systemd-pcrlock cel | jq --argjson current "$(cat /var/lib/pcrlock.d/500-separator.pcrlock.d/generated.pcrlock)" '
    [
        .[]
        | select(.pcr == 7 and .content.event_type == "EV_SEPARATOR")
        | with_entries(select([.key] | inside(["pcr", "digests"])))
    ] as $newRecord
    | $current
    | .records |= [ .[] | select(.pcr != 7), $newRecord ]
' >/var/lib/pcrlock.d/500-separator.pcrlock.d/generated.pcrlock.new
mv /var/lib/pcrlock.d/500-separator.pcrlock.d/generated.pcrlock.new /var/lib/pcrlock.d/500-separator.pcrlock.d/generated.pcrlock

# After separator.
/run/current-system/systemd/lib/systemd/systemd-pcrlock cel | jq '
    [.[] | select(.pcr == 7)] as $pcr7 
    | ($pcr7.[] | select(.content.event_type == "EV_SEPARATOR")) as $sep 
    | { "records": [
        $pcr7 | .[index($sep)+1:]
        | .[] | select(.content.event_type == "EV_EFI_VARIABLE_AUTHORITY")
        | with_entries(select([.key] | inside(["pcr", "digests"])))
    ]}
' >/var/lib/pcrlock.d/620-secureboot-authority.pcrlock.d/generated.pcrlock
```

docs/how-to-guides/enable-measured-boot.md
```diff
@@ -22,6 +22,27 @@ If this says anything other than `yes`, you will not be able to use Lanzaboote
 for Measured Boot because the TPM in your system is not supported by
 systemd-pcrlock.
 
+### Check PCR 7 Ordering
```
nikstur, line 25: I think this is too much of a niche problem and also not really actionable for users and thus doesn't belong in the guide.

Please move this to troubleshooting and then you can link to the troubleshooting section from the guide.

It would also be nice if you could describe the failure you would see when you try to follow the guide but this doesn't work.

nikstur (changes requested):
Thank your for this contribution!

Issue #614, opened 2026-06-04
Measure Boot: Always keep currently booted measurements


giggio:
I was not sure what was triggering the error, it happens intermittently, and it does not happen all the time. But I got an edge case here.

I've been changing files that will end up initrd, which trigger new measurements. I do it with `nixos-rebuild switch` and it works. I get the expected message: `PCR 4 (boot-loader-code) matches event log and fully consists of recognized measurements. Including in set of PCRs.`
PCR 4 is still included in `/var/lib/systemd/pcrlock.json` and after a reboot it is still there.
But after I've done this a few times (greater than `configurationLimit`) without rebooting, I start getting:

```
PCR 4 (boot-loader-code) event log contains unrecognized measurements. Removing from set of PCRs.
```

Then, it doesn't matter how many times I retry, it does not work anymore. Not until I run `systemd-pcrlock remove-policy && nixos-rebuild boot` and reboot.
This obviously messes up my LUKS volume TPM unlock policy, and I have to wipe and reenroll.
When I run `systemd-pcrlock remove-policy && nixos-rebuild boot` I still get the error about PCR 4 having unrecognized mearuments, and `/var/lib/systemd/pcrlock.json` does not have PCR 4 in it, but after rebooting it does have PCR 4 in it and everthing is fine, until it happens again.

What seems to me to be happening is that after the 8th time (as per `configurationLimit`) you update the boot files the `.pcrlock` that locked your current boot is removed from `/var/lib/pcrlock.d/635-lanzaboote.pcrlock.d/`, and then the prediction for PCR 4 fails and is not included anymore, not until you remove the policy, run a `nixos-rebuild boot` and reboot.

So, it would seem apropriate to not remove any file that is currently being used as a component from `/var/lib/pcrlock.d/635-lanzaboote.pcrlock.d/`, maybe remove all the ones up until the one in use, but not the one in use, and any one after it.

nikstur:
> So, it would seem apropriate to not remove any file that is currently being used as a component from /var/lib/pcrlock.d/635-lanzaboote.pcrlock.d/, maybe remove all the ones up until the one in use, but not the one in use, and any one after it.

This sounds reasonable to me.

WilliButz:
[...]

PR #637, opened 2026-07-19
tool: make Measured Boot policy cover PCR4 throughout bootloader update

WilliButz:
Relates to #614 

Before this change, a bootloader update would cause PCR4 to no longer be part of the generated pcrlock policy because there only was a component file (aka. measurement) for the freshly installed, but not yet booted bootloader, so the policy generated from the components could not match the PCR4 state of the TPM.

This meant that the policy was much less strict between this bootloader update and the next rebuild *after* a reboot that includes the new bootloader, therby undermining the main goal that the Measured Boot integration tries to achieve.

Now, the component file for the bootloader from before the update is kept so that that the policy continues to cover PCR4. It is important for users to reboot between two bootloader updates, otherwise the issue above arises again. Unfortunately that case cannot be easily addressed without much more involved logic in lanzaboote.

A downside from this is that the maximum number of NixOS generations is further reduced from 8 to 4, because we now always have two variants for the bootloader. This is nothing lanzaboote can work around in any meaningful way, instead this needs to be addressed in systemd. See the upstream issue: https://github.com/systemd/systemd/issues/41526

### DIFF
--- a/nix/modules/lanzaboote.nix
+++ b/nix/modules/lanzaboote.nix
@@ -429,13 +429,13 @@ in
         '';
       }
       {
-        assertion = cfg.measuredBoot.enable -> (configurationLimit > 0 && configurationLimit <= 8);
+        assertion = cfg.measuredBoot.enable -> (configurationLimit > 0 && configurationLimit <= 4);
         message = ''
-          If Measured Boot is enabled, you cannot store more than 8 generations on the ESP.
+          If Measured Boot is enabled, you cannot store more than 4 generations on the ESP.
 
-            This is a strict limit required and enforced by systemd-pcrlock.
+            This is a strict limit introduced and enforced by systemd-pcrlock.
 
-            Set `boot.lanzaboote.configurationLimit = 8;` to reduce the number of generations you store.
+            Set `boot.lanzaboote.configurationLimit = 4;` to reduce the number of generations you store.
         '';
       }
     ];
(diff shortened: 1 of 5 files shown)


rust/tool/systemd/src/install.rs
```diff
@@ -466,42 +466,44 @@ impl<S: Signer> Installer<S> {
...
-                // bootloader version installed on the ESP and there is no rollback mechanism for it.
-                // That's why we also do not extend the GC roots with the path.
-                //
-                // However, we call this file "next" so that while the new bootloader is installed,
-                // the measurements of the current and next remain available.
-                let bootloader_pcrlock = pcrlock_paths.bootloader_measurement("next");
-                lock_pe(&systemd_boot, &bootloader_pcrlock)
-                    .context("Failed to lock Lanzaboote image with systemd-pcrlock")?;
+                // Path to measurement that was used prior to this implementation
+                let legacy_path = pcrlock_paths.bootloader_measurement("next");
```
nikstur, line 477: I would prefer if we just delete "next" if it exists. Then we can eventually just remove this code path after a few releases.

nikstur (approved):
Awesome, thank you!

nix/modules/lanzaboote.nix
```diff
@@ -429,13 +429,13 @@ in
         '';
       }
       {
-        assertion = cfg.measuredBoot.enable -> (configurationLimit > 0 && configurationLimit <= 8);
+        assertion = cfg.measuredBoot.enable -> (configurationLimit > 0 && configurationLimit <= 4);
```
SolidEva, line 432: sorry for the drive by, but might need to update the docs https://github.com/nix-community/lanzaboote/blob/c191775376b8734af02970153b4e5d86841dff19/docs/how-to-guides/enable-measured-boot.md?plain=1#L28

